PR #13, opened 2020-02-05
Updates and fixes for unit tests

froschdesign:
|    Q          |   A
|-------------- | ------
| QA            | yes


test/Writer/Extension/ITunes/FeedTest.php
```diff
@@ -209,8 +209,8 @@ public function testSetKeywordsThrowsExceptionIfMaxKeywordsExceeded()
         }, \E_USER_DEPRECATED);
         try {
             $feed->setItunesKeywords($words);
-            $this->fail('Expected exception when setting more keywords than allowed');
         } catch (ExceptionInterface $e) {
+            $this->assertInstanceOf(ExceptionInterface::class, $e);
```
michalbundyra: I am not sure about this change. In case there will be no any exception the test will pass. Right?

test/Writer/FeedTest.php
```diff
@@ -63,45 +63,54 @@ public function testAddsAuthorUriFromArray()
 
     public function testAddAuthorThrowsExceptionOnInvalidNameFromArray()
     {
+        $this->expectException(
+            Writer\Exception\InvalidArgumentException::class
+        );
```
michalbundyra: minor: should be in one line, imo

test/Writer/FeedTest.php
```diff
@@ -63,45 +63,54 @@ public function testAddsAuthorUriFromArray()
...
-                                     'email' => '']);
-            $this->fail();
-        } catch (Writer\Exception\InvalidArgumentException $e) {
-        }
+        $writer->addAuthor(
+            [
+                'name'  => 'Joe',
+                'email' => '',
+            ]
+        );
```
michalbundyra: minor: for calls like that I'd prefer:
```php
$something->foo([
    'bar' => 'baz',
    'abc' => 'def',
]);
```

test/Writer/FeedTest.php
```diff
@@ -654,12 +638,11 @@ public function testSetsGeneratorThrowsExceptionOnInvalidVersionDeprecated()
     public function testSetsGeneratorThrowsExceptionOnInvalidUriDeprecated()
     {
         $this->markTestIncomplete('Pending Laminas\URI fix for validation');
```
michalbundyra: out of scope,  but maybe this is already fixed, and we can remove above?

test/Writer/Renderer/Entry/AtomTest.php
```diff
@@ -62,7 +62,8 @@ public function tearDown()
     public function testRenderMethodRunsMinimalWriterContainerProperlyBeforeICheckAtomCompliance()
     {
         $renderer = new Renderer\Feed\Atom($this->validWriter);
-        $renderer->render();
+
+        $this->assertInstanceOf(Renderer\Feed\Atom::class, $renderer->render());
```
michalbundyra: I am not sure about this test, we tend to drop fluent interfaces, so we would need to change it back when we do it.
Or then we will use `assertNull`?

I know, that this is reported as "risky" because of no assertion.

test/Writer/Extension/ITunes/EntryTest.php
```diff
@@ -158,8 +158,8 @@ public function testSetKeywordsThrowsExceptionIfMaxKeywordsExceeded()
         }, \E_USER_DEPRECATED);
         try {
             $entry->setItunesKeywords($words);
-            $this->fail('Expected exception when setting more keywords than allowed');
         } catch (ExceptionInterface $e) {
+            $this->assertInstanceOf(ExceptionInterface::class, $e);
```
Xerkus: Use `expectException()` and omit catch block entirely, same for the rest.

test/Writer/FeedTest.php
Xerkus: I agree, when it fits within line limit. Up to personal preference imo.

test/Writer/Extension/ITunes/FeedTest.php
froschdesign: We need this order here to ensure the restore of the error handler.

test/Writer/Renderer/Entry/AtomTest.php
Xerkus: When the lack of exception is the assertion, `$this->addToAssertionCount(1);` is a good choice instead of meaningless or unrelated assertion.

froschdesign: The entire API is not good and it will change with the next major version but that is not the goal of this pull request..

test/Writer/Extension/ITunes/FeedTest.php
michalbundyra: My point is that we need some sort of fail in case `setItunesKeywords` does not throw any exception.

And I believe finally will be executed anyway - I mean if we keep `$this->fail` call in `try`.

test/Writer/Renderer/Entry/AtomTest.php
Xerkus: Tangential: This test does not assert what the name states imo. Test name needs to be amended or test updated.

test/Writer/FeedTest.php
froschdesign: I am still waiting for the coding standard documentation. 😉

test/Writer/Extension/ITunes/EntryTest.php
froschdesign: Please give me a hint how to restore the error handler then.

test/Writer/Renderer/Entry/AtomTest.php
michalbundyra: I do not agree that `addToAssertionCount` is better in that case. This does not give us any information at all.
I've checked it shortly and people wanted something like "assertNotThrowException" but it never happened in PHPUnit. Assertion on the result of the method is the best imo. If we drop fluent interfaces (in next major) we will change it to `assertNull` and all will be fine. At least for now there is no distinction in null and void in PHP ;-)

test/Writer/Extension/ITunes/EntryTest.php
Xerkus: @froschdesign I didn't say try-catch, only catch. It is optional in the presence of finally

test/Writer/Renderer/Entry/AtomTest.php
Xerkus: I strongly disagree. Return type is beside the point for the test. If it changes it would not change the validity of the test outcome. More than that, asserting return value in any form actually distracts from the actual implicit test condition. It is better to show that we have no explicit assertions and extra attention should be paid to the test.

michalbundyra: I think we are looking for [`@doesNotPerformAssertions`](https://phpunit.readthedocs.io/en/8.5/annotations.html#doesnotperformassertions) here.

But personally, as I said, I would go with assertion of the result. I know it tests two things, and we have "single assertion tests" but it's better that tricking PHPUnit that we have assertion, once we haven't.

Xerkus: @michalbundyra That annotation was first implemented in 5.6.0. I guess I can start using it already :joy:

test/Writer/FeedTest.php
froschdesign: No we can not, because there are no tests for `Laminas\Feed\Uri` and so we know nothing.

Example:

```php
var_dump(Laminas\Feed\Uri::factory('notauri')->isValid()); // true
```

test/Writer/EntryTest.php
```diff
@@ -62,45 +62,48 @@ public function testAddsAuthorUriFromArray()
...
-            $this->fail();
-        } catch (Writer\Exception\InvalidArgumentException $e) {
-        }
+        $entry->addAuthor(
+            [
+                'name'  => 'Joe',
+                'email' => 'joe@example.org',
+                'uri'   => '',
+            ]
+        );
```
Xerkus: Please
```suggestion
        $entry->addAuthor([
            'name'  => 'Joe',
            'email' => 'joe@example.org',
            'uri'   => '',
        ]);
```

Xerkus: Please reformat for better readability:
```php
        $entry->addAuthor([
            'name'  => 'Joe',
            'email' => 'joe@example.org',
            'uri'   => '',
        ]);
```

test/Writer/FeedTest.php
```diff
@@ -559,35 +532,38 @@ public function testSetsGeneratorUri()
...
-                                        'uri' => 'notauri']);
-            $this->fail();
-        } catch (ExceptionInterface $e) {
-        }
+        $writer->setGenerator(
+            [
+                'name' => 'LaminasW',
+                'uri'  => 'notauri',
+            ]
+        );
```
Xerkus:
```php
        $writer->setGenerator([
            'name' => 'LaminasW',
            'uri'  => 'notauri',
        ]);
```

test/Writer/FeedTest.php
```diff
@@ -777,24 +740,28 @@ public function testAddsManyCategories()
...
-                                       'scheme' => 'http://www.example.com/schema1']);
-            $this->fail();
-        } catch (ExceptionInterface $e) {
-        }
+        $writer->addCategory(
+            [
+                'label'  => 'Cats & Dogs',
+                'scheme' => 'http://www.example.com/schema1',
+            ]
+        );
```
Xerkus:
```php
        $writer->addCategory([
            'label'  => 'Cats & Dogs',
            'scheme' => 'http://www.example.com/schema1',
        ]);
```

test/Writer/FeedTest.php
```diff
@@ -777,24 +740,28 @@ public function testAddsManyCategories()
...
-                                       'scheme' => 'http://']);
-            $this->fail();
-        } catch (ExceptionInterface $e) {
-        }
+        $writer->addCategory(
+            [
+                'term'   => 'cat_dog',
+                'scheme' => 'http://',
+            ]
+        );
```
Xerkus:
```php
        $writer->addCategory([
            'term'   => 'cat_dog',
            'scheme' => 'http://',
        ]);
```

test/Writer/Renderer/Entry/RssTest.php
```diff
@@ -253,36 +256,51 @@ public function testAddsEnclosureThrowsExceptionOnNegativeLength()
...
-                                              'uri'    => 'http://example.com/audio.mp3',
-                                              'length' => 0
-                                         ]);
+        $this->validEntry->setEnclosure(
+            [
+                'type'   => 'audio/mpeg',
+                'uri'    => 'http://example.com/audio.mp3',
+                'length' => 0,
+            ]
+        );
```
Xerkus:
```php
        $this->validEntry->setEnclosure([
            'type'   => 'audio/mpeg',
            'uri'    => 'http://example.com/audio.mp3',
            'length' => 0,
        ]);
```

test/Writer/Renderer/Entry/RssTest.php
```diff
@@ -253,36 +256,51 @@ public function testAddsEnclosureThrowsExceptionOnNegativeLength()
...
-                                              'uri'    => 'http://example.com/audio.mp3',
-                                              'length' => 23
-                                         ]);
+        $this->validEntry->setEnclosure(
+            [
+                'type'   => 'audio/mpeg',
+                'uri'    => 'http://example.com/audio.mp3',
+                'length' => 23,
+            ]
+        );
```
Xerkus:
```php
        $this->validEntry->setEnclosure([
            'type'   => 'audio/mpeg',
            'uri'    => 'http://example.com/audio.mp3',
            'length' => 23,
        ]);
```

test/Writer/Renderer/Entry/RssTest.php
```diff
@@ -253,36 +256,51 @@ public function testAddsEnclosureThrowsExceptionOnNegativeLength()
...
-                                              'uri'    => 'http://example.com/audio.mp3',
-                                              'length' => '23'
-                                         ]);
+        $this->validEntry->setEnclosure(
+            [
+                'type'   => 'audio/mpeg',
+                'uri'    => 'http://example.com/audio.mp3',
+                'length' => '23',
+            ]
+        );
```
Xerkus:
```php
        $this->validEntry->setEnclosure([
            'type'   => 'audio/mpeg',
            'uri'    => 'http://example.com/audio.mp3',
            'length' => '23',
        ]);
```

test/Writer/Extension/ITunes/EntryTest.php
```diff
@@ -158,15 +160,16 @@ public function testSetKeywordsThrowsExceptionIfMaxKeywordsExceeded()
         }, \E_USER_DEPRECATED);
         try {
             $entry->setItunesKeywords($words);
-            $this->fail('Expected exception when setting more keywords than allowed');
-        } catch (ExceptionInterface $e) {
+            $this->fail();
```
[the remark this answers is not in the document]
froschdesign: https://github.com/laminas/laminas-feed/pull/13#discussion_r375701304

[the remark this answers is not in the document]
Xerkus: @froschdesign If exception is not thrown then test will fail because exception is expected and expectation is not met

froschdesign:
@Xerkus @michalbundyra 
Please take a look at the whole file and all the others and you will see that the problems with the coding style are everywhere.
And you are right we must fix these problems but this can be done later per automatism.

It would be great if you can help me here with the tests:

```
1) LaminasTest\Feed\PubSubHubbub\Subscriber\CallbackTest::testCanSetHttpResponseObject
Function ReflectionType::__toString() is deprecated
```

https://travis-ci.com/laminas/laminas-feed/jobs/284472947#L399-L400

https://github.com/laminas/laminas-feed/blob/42ca5e389f6b66e38fe58f349af676c54fbc5e52/test/PubSubHubbub/Subscriber/CallbackTest.php#L497-L516

Xerkus:
I agree about whole thing being inconsistent but CS changes are requested only for the stuff you touched.

> It would be great if you can help me here with the tests:
> 
> ```
> 1) LaminasTest\Feed\PubSubHubbub\Subscriber\CallbackTest::testCanSetHttpResponseObject
> Function ReflectionType::__toString() is deprecated
> ```

phphunit 7 or above is needed for php 7.4:
https://travis-ci.com/laminas/laminas-feed/jobs/284472947#L333

froschdesign:
@Xerkus 
> I agree about whole thing being inconsistent but CS changes are requested only for the stuff you touched.

Can all be done later.

> phphunit 7 or above is needed for php 7.4:

Right, but can you help to use Prophecy here? Unfortunately I have not found a suitable solution.
It already works for [small fixes](https://github.com/laminas/laminas-feed/pull/13/commits/603d4d2528c4da7b545a47c1f17807ffbea6af12).

.travis.yml
```diff
@@ -45,6 +45,12 @@ matrix:
     - php: 7.3
       env:
         - DEPS=latest
+    - php: 7.4
+      env:
+        - DEPS=lowest
```
Xerkus: Lowest can not be used on 7.4 at this time as the phpunit below 7 will cause deprecation warnings.
One option is to do `composer update phpunit/phpunit --update-with-dependencies`, however in other repositories we opted to omit 7.4 lowest until unsupported php versions are dropped and phpunit bumped to 8.

Xerkus:
> Right, but can you help to use Prophecy here? Unfortunately I have not found a suitable solution.
> It already works for [small fixes](https://github.com/laminas/laminas-feed/pull/13/commits/603d4d2528c4da7b545a47c1f17807ffbea6af12).

Is there any specific reason to use prophecy? Mock builder is not deprecated afaik.

froschdesign:
> Is there any specific reason to use prophecy? Mock builder is not deprecated afaik.

The usage of Prophecy avoids the warnings like:

```
LaminasTest\Feed\PubSubHubbub\Subscriber\CallbackTest::testCanSetHttpResponseObject
Function ReflectionType::__toString() is deprecated
```

Xerkus:
But then updating phpunit to a version that supports  7.4 does too

Xerkus:
Also add phpunit 7
https://github.com/laminas/laminas-feed/blob/3a867ea44130eb79646367fecbc556e8d966afae/composer.json#L42
```
        "phpunit/phpunit": "^5.7.23 || ^6.4.3 || ^7.5.20",

File: test/Writer/EntryTest.php [hunk 16]
```diff
@@ -477,35 +458,42 @@ public function testSetsCommentFeedLink()
...
-                                             'type' => 'rdf']);
-            $this->fail();
-        } catch (Writer\Exception\ExceptionInterface $e) {
-        }
+        $entry->setCommentFeedLink(
+            [
+                'uri'  => 'http://',
+                'type' => 'rdf',
+            ]
+        );
```
Xerkus:
```php
        $entry->setCommentFeedLink([
            'uri'  => 'http://',
            'type' => 'rdf',
        ]);
```

test/Writer/EntryTest.php
```diff
@@ -477,35 +458,42 @@ public function testSetsCommentFeedLink()
...
-                                             'type' => 'foo']);
-            $this->fail();
-        } catch (Writer\Exception\ExceptionInterface $e) {
-        }
+        $entry->setCommentFeedLink(
+            [
+                'uri'  => 'http://www.example.com/id/comments',
+                'type' => 'foo',
+            ]
+        );
```
Xerkus:
```php
        $entry->setCommentFeedLink([
            'uri'  => 'http://www.example.com/id/comments',
            'type' => 'foo',
        ]);
```

test/Writer/EntryTest.php
```diff
@@ -477,35 +458,42 @@ public function testSetsCommentFeedLink()
...
-                                             'type' => 'rdf']);
-            $this->fail();
-        } catch (Writer\Exception\ExceptionInterface $e) {
-        }
+        $entry->setCommentFeedLink(
+            [
+                'uri'  => '',
+                'type' => 'rdf',
+            ]
+        );
```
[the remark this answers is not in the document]
froschdesign: We definitely need an automated solution otherwise we generate too much noise in the commit history and scare away every developer.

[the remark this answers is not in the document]
Xerkus: There is a great tool in git for keeping history clean but it is vastly underused:
```
git commit --fixup <hash of commit being fixed>
```
Once review rounds are complete:
```
git rebase -i --autosquash <PR base  branch>
```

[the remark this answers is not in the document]
froschdesign: An automated "coding standard fixer" helps everyone and saves time for all sides.

michalbundyra:
Thanks, @froschdesign!

Issue #30, opened 2020-11-20
PHPDoc return types incorrect/outdated


Synchro:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | dev-master (currently 2.14.x)

#### Summary

Return types declared in phpdocs are often incorrect, probably because they are outdated.

For example, in [laminas-feed/src/Reader/Entry/Rss.php](https://github.com/laminas/laminas-feed/blob/2.14.x/src/Reader/Entry/Rss.php#L278) the `getEnclosure()` method is declared as:

    @return string

when it actually returns a `stdClass` instance. This is enough to trip up most IDEs and static analysers. I think in this particular case it used to return an array instead of a stdClass, but string is wrong for that too.

There are many examples of this, and it probably applies to other components too. Because it's so widespread, I didn't want to do a PR on spec in case you have some policy on such phpdocs, but I wanted to flag it anyway.

froschdesign:
@Synchro 
> Return types declared in phpdocs are often incorrect, probably because they are outdated.

You are right. In many cases they are completely wrong. See an example in https://github.com/laminas/laminas-feed/pull/33#issuecomment-750874854

We have to rework the entire component with all classes and their interfaces.

PR #33, opened 2020-12-20
Fixes 32 - Method getDateModified of RSS reader doesn't iterate over different formats

### LINKED ISSUE #32: Method getDateModified of RSS reader doesn't iterate over different formats
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.13.0

#### Summary

The current code suggest that when retrieving the date modified variable it will attempt to create it from 3 different formats: `RSS`, `RFC822` and `RFC2822`. This doesn't happen because the method used `date_create_from_format` doesn't throw an exception.
This also means that if the format is not recognised at all the exception won't be thrown.

#### Current behavior

If the given date doesn't match the first format tested (`RSS`) the variable will resolve into false and the loop will break there.

#### How to reproduce

Read an RSS feed with a date modified format different from `RSS` (`D, d M Y H:i:s O`).

#### Expected behavior

The code will iterate over all 3 different formats and if unable to create the datetime object it will throw an exception saying the date format was not recognised.

For reference, the piece of code with the bug:
https://github.com/laminas/laminas-feed/blob/2.13.x/src/Reader/Feed/Rss.php#L205

```
$dateStandards = [
    DateTime::RSS,
    DateTime::RFC822,
    DateTime::RFC2822,
    null,
];

foreach ($dateStandards as $standard) {
    try {
        $date = date_create_from_format($standard, $dateModified);
        break;
    } catch (\Exception $e) {
        if ($standard === null) {
            throw new Exception\RuntimeException(
                'Could not load date due to unrecognised format'
                . ' (should follow RFC 822 or 2822): ' . $e->getMessage(),
                0,
                $e
            );
        }
    }
}
```

froschdesign:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### DIFF
--- a/src/Reader/Feed/Rss.php
+++ b/src/Reader/Feed/Rss.php
@@ -206,22 +206,21 @@ class Rss extends AbstractFeed
                         DateTime::RSS,
                         DateTime::RFC822,
                         DateTime::RFC2822,
-                        null,
                     ];
                     foreach ($dateStandards as $standard) {
-                        try {
-                            $date = DateTime::createFromFormat($standard, $dateModified);
+                        $date = DateTime::createFromFormat(
+                            $standard,
+                            $dateModified
+                        );
+                        if ($date instanceof DateTime) {
                             break;
-                        } catch (\Exception $e) {
-                            if ($standard === null) {
-                                throw new Exception\RuntimeException(
-                                    'Could not load date due to unrecognised format'
-                                    . ' (should follow RFC 822 or 2822): ' . $e->getMessage(),
-                                    0,
-                                    $e
-                                );
-                            }
                         }
+                    }
+                    if (! $date) {
+                        throw new Exception\RuntimeException(
+                            'Could not load date due to unrecognised'
+                            . ' format (should follow RFC 822 or 2822).'
+                        );
                     }
                 }
             }
--- a/test/Reader/Feed/RssTest.php
+++ b/test/Reader/Feed/RssTest.php
@@ -19,6 +19,7 @@ use PHPUnit\Framework\TestCase;
  */
 class RssTest extends TestCase
 {
+    /** @var string */
     protected $feedSamplePath;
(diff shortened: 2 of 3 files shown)


weierophinney (changes requested):
Changes look reasonable to me. Please check the Psalm output, though, as you'll either need to correct some new issues, or add to the baseline:

- https://travis-ci.com/github/laminas/laminas-feed/jobs/464005564#L793

froschdesign:
@weierophinney 
> Please check the Psalm output, though, as you'll either need to correct some new issues…

Yes, there are massive problems with many bugs. But these are _old_ problems and not related to this pull request. No new problems are introduced with this PR.

---

Pslam reports the following problem for example:

> RedundantCastGivenDocblockType - test/Reader/Entry/AtomStandaloneEntryTest.php:150:39
>
> Redundant cast to array given docblock-provided type (see https://psalm.dev/263)
> $this->assertEquals($authors, (array) $entry->getAuthors());

https://travis-ci.com/github/laminas/laminas-feed/jobs/464005564#L881-L882

But the problem is not the redundant cast, the problem is the related DocBlock:

https://github.com/laminas/laminas-feed/blob/db2d64788a25ab10b77c6301b42692b2953bcc3c/src/Reader/Feed/FeedInterface.php#L26-L31

It's not a array it's a collection or `null`:

https://github.com/laminas/laminas-feed/blob/db2d64788a25ab10b77c6301b42692b2953bcc3c/src/Reader/Collection/Author.php#L11-L27

https://github.com/laminas/laminas-feed/blob/db2d64788a25ab10b77c6301b42692b2953bcc3c/src/Reader/Feed/Rss.php#L124-L126

froschdesign:
There is a separate issue report for the wrong DocBlocks: #30

weierophinney (approved):

PR #35, opened 2021-02-08
Switch from Travis-CI to GitHub Actions

weierophinney:
Adopts workflow prototyped in laminas/laminas-eventmanager#15 - mainly to check that Psalm integration works as expected.


boesing:
Is psalm ignoring baseline in the GHA?

weierophinney:
> Is psalm ignoring baseline in the GHA?

Part of it, but not all of it; my guess is a difference in Psalm versions between last run and this one — which is why @Ocramius is suggesting we always commit a lockfile, to ensure tooling runs on consistent versions.

PR #39, opened 2021-05-10
Remove file headers

ghostwriter:
**[QA]** Remove file header (DocBlock comments)


Ocramius:
Done as part of #42 - thanks for the patch nonetheless, @ghostwriter! I've included it in patch notes :)

PR #42, opened 2021-09-20
Provide PHP 8.1 support

weierophinney:
This patch prepares the component for use with PHP 8.1, via the following actions:

- Adds `~8.1.0` to the list of allowed PHP versions
- Changes how the package replaces zend-feed:
  - Removes the "replace" section of the package definition
  - Adds `"zendframework/zend-feed": "*"` to the list of package conflicts
- Bumps laminas-coding-standard to the 2.2 ruleset and applies it
- Bumps dependency versions to those known to work with 8.1 when possible
- Ensures all functionality works under 8.1, primarily by adding `#[ReturnTypeWillChange]` attributes to `Iterator`, `Countable`, and `ArrayObject` method implementations.

### DIFF
--- a/src/PubSubHubbub/Subscriber.php
+++ b/src/PubSubHubbub/Subscriber.php
@@ -1,19 +1,31 @@
 <?php
 
-/**
- * @see       https://github.com/laminas/laminas-feed for the canonical source repository
- * @copyright https://github.com/laminas/laminas-feed/blob/master/COPYRIGHT.md
- * @license   https://github.com/laminas/laminas-feed/blob/master/LICENSE.md New BSD License
- */
-
 namespace Laminas\Feed\PubSubHubbub;
 
 use DateInterval;
 use DateTime;
 use Laminas\Feed\Uri;
+use Laminas\Http\Client;
 use Laminas\Http\Request as HttpRequest;
 use Laminas\Stdlib\ArrayUtils;
 use Traversable;
+
+use function array_key_exists;
+use function array_search;
+use function array_unique;
+use function gettype;
+use function hash;
+use function implode;
+use function in_array;
+use function intval;
+use function is_array;
+use function is_string;
+use function md5;
+use function rand;
+use function rtrim;
+use function time;
+use function uksort;
+use function uniqid;
 
 class Subscriber
 {
@@ -298,7 +310,8 @@ class Subscriber
      */
     public function setPreferredVerificationMode($mode)
     {
-        if ($mode !== PubSubHubbub::VERIFICATION_MODE_SYNC
+        if (
+            $mode !== PubSubHubbub::VERIFICATION_MODE_SYNC
             && $mode !== PubSubHubbub::VERIFICATION_MODE_ASYNC
         ) {
             throw new Exception\InvalidArgumentException(
@@ -435,7 +448,7 @@ class Subscriber
     /**
      * Add an optional parameter to the (un)subscribe requests
      *
-     * @param  string      $name
+     * @param  string|array<string, string> $name
      * @param  null|string $value
      * @return $this
      * @throws Exception\InvalidArgumentException
@@ -593,6 +606,8 @@ class Subscriber
         return $this->asyncHubs;
     }
 
+    // phpcs:disable PSR2.Methods.MethodDeclaration.Underscore
+
     /**
      * Executes an (un)subscribe request
      *
@@ -600,10 +615,8 @@ class Subscriber
      * @return void
      * @throws Exception\RuntimeException
      */
-    // @codingStandardsIgnoreStart
     protected function _doRequest($mode)
     {
-        // @codingStandardsIgnoreEnd
         $client = $this->_getHttpClient();
         $hubs   = $this->getHubUrls();
         if (empty($hubs)) {
@@ -619,9 +632,10 @@ class Subscriber
                 $client->setAuth($auth[0], $auth[1]);
             }
             $client->setUri($url);
-            $client->setRawBody($params = $this->_getRequestParameters($url, $mode));
+            $client->setRawBody($this->_getRequestParameters($url, $mode));
             $response = $client->send();
-            if ($response->getStatusCode() !== 204
+            if (
+                $response->getStatusCode() !== 204
                 && $response->getStatusCode() !== 202
             ) {
                 $this->errors[] = [
@@ -636,7 +650,7 @@ class Subscriber
              * are using async verification modes so they may update Models and
              * move these to asynchronous processes.
              */
-            } elseif ($response->getStatusCode() == 202) {
+            } elseif ($response->getStatusCode() === 202) {
                 $this->asyncHubs[] = [
                     'response' => $response,
                     'hubUrl'   => $url,
@@ -648,12 +662,10 @@ class Subscriber
     /**
      * Get a basic prepared HTTP client for use
      *
-     * @return \Laminas\Http\Client
+     * @return Client
      */
-    // @codingStandardsIgnoreStart
     protected function _getHttpClient()
     {
-        // @codingStandardsIgnoreEnd
         $client = PubSubHubbub::getHttpClient();
         $client->setMethod(HttpRequest::METHOD_POST);
         $client->setOptions([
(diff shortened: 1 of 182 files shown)


src/PubSubHubbub/Subscriber.php
```diff
@@ -735,19 +745,22 @@ protected function _getRequestParameters($hubUrl, $mode)
...
-            'expiration_time' => $expires,
-            // @codingStandardsIgnoreStart
-            'subscription_state' => ($mode == 'unsubscribe') ? PubSubHubbub::SUBSCRIPTION_TODELETE : PubSubHubbub::SUBSCRIPTION_NOTVERIFIED,
-            // @codingStandardsIgnoreEnd
+            'id'            => $key,
+            'topic_url'     => $params['hub.topic'],
+            'hub_url'       => $hubUrl,
+            'created_time'  => $now->format('Y-m-d H:i:s'),
+            'lease_seconds' => $params['hub.lease_seconds'],
+            /** @psalm-suppress PossiblyInvalidCast */
```
Ocramius, line 756: Seems fishy - `$params['hub.verify_token']` always contains a `string` at this location?

Ocramius: If not, a comment is needed - can't wrap my mind around this comment set here as standalone

Ocramius: Specifically, this is generated by:

```php
    protected function _generateVerifyToken()
    {
        // @codingStandardsIgnoreEnd
        if (! empty($this->testStaticToken)) {
            return $this->testStaticToken;
        }
        return uniqid(rand(), true) . time();
    }
```

src/Reader/AbstractFeed.php
```diff
@@ -259,14 +265,15 @@ public function __call($method, $args)
      * Return an Extension object with the matching name (postfixed with _Feed)
      *
      * @param  string $name
-     * @return Extension\AbstractFeed
+     * @return null|Extension\AbstractFeed
```
Ocramius, line 268: This change is a bit suspicious :thinking:

src/Reader/Entry/AbstractEntry.php
```diff
@@ -173,14 +175,15 @@ public function getExtensions()
      * Return an Extension object with the matching name (postfixed with _Entry)
      *
      * @param  string $name
-     * @return Reader\Extension\AbstractEntry
+     * @return null|Reader\Extension\AbstractEntry
```
Ocramius, line 178: This change is a bit suspicious :thinking:

src/Reader/Entry/Atom.php
```diff
@@ -70,6 +69,7 @@ public function getAuthors()
             return $this->data['authors'];
         }
 
+        /** @psalm-suppress PossiblyNullReference */
```
Ocramius: We should probably use a baseline, instead of suppressing these? Should we perhaps assert that the extension got instantiated instead? (applies to subsequent `->getExtension()` calls too)

Ocramius: Perhaps isolating to `private function getAtomExtension()`?

src/Reader/Entry/Rss.php
```diff
@@ -84,7 +89,9 @@ public function getAuthors()
             return $this->data['authors'];
         }
 
-        $authors   = [];
+        $authors = [];
+
+        /** @psalm-suppress PossiblyNullReference */
```
Ocramius: Similar to other `getExtension()` usages, these suppressions should IMO be removed and/or isolated to a new `private function getDublinCoreExtension()`

src/Reader/Entry/Rss.php
```diff
@@ -562,13 +598,16 @@ public function getCommentFeedLink()
...
         $commentfeedlink = $this->getExtension('WellFormedWeb')->getCommentFeedLink();
 
         if (! $commentfeedlink) {
+            /** @psalm-suppress PossiblyNullReference */
             $commentfeedlink = $this->getExtension('Atom')->getCommentFeedLink('rss');
         }
 
         if (! $commentfeedlink) {
+            /** @psalm-suppress PossiblyNullReference */
```
Ocramius: Overall, we really need to get rid of these - they are everywhere @_@

src/Reader/Extension/Atom/Feed.php
```diff
@@ -166,6 +166,7 @@ public function getDescription()
             return $this->data['description'];
         }
 
+        /** @psalm-suppress UnusedVariable */
```
Ocramius: This variable **is** unused though?

src/Reader/Extension/Atom/Feed.php
```diff
@@ -82,6 +81,7 @@ public function getCopyright()
             return $this->data['copyright'];
         }
 
+        /** @psalm-suppress UnusedVariable */
```
Ocramius: This variable **is** unused though?

src/Uri.php
```diff
@@ -178,6 +160,6 @@ public function isValid()
      */
     public function isAbsolute()
     {
-        return $this->scheme !== null;
+        return ! empty($this->scheme);
```
Ocramius, line 163: `! empty()` here seems wrong, considering `$this->scheme` is `string|null`

src/Writer/ExtensionPluginManager.php
```diff
@@ -22,29 +22,30 @@ class ExtensionPluginManager extends AbstractPluginManager implements ExtensionM
     /**
      * Aliases for default set of extension classes
      *
-     * @var array
+     * @var array<array-key, string>
```
Ocramius, line 25: `array<string, ...` rather than `array<array-key, `

test/PubSubHubbub/PublisherTest.php
```diff
@@ -314,7 +309,8 @@ public function testNotifiesHubAndReportsSuccess(): void
     public function testNotifiesHubAndReportsFail(): void
     {
         PubSubHubbub::setHttpClient($this->getClientFail());
-        $client = PubSubHubbub::getHttpClient();
+        PubSubHubbub::getHttpClient();
```
Ocramius: Unused?

test/PubSubHubbub/PublisherTest.php
```diff
@@ -303,7 +297,8 @@ public function testNotifiesHubWithCorrectParametersAndMultipleTopics(): void
     public function testNotifiesHubAndReportsSuccess(): void
     {
         PubSubHubbub::setHttpClient($this->getClientSuccess());
-        $client = PubSubHubbub::getHttpClient();
+        PubSubHubbub::getHttpClient();
```
Ocramius: Unused?

Ocramius (changes requested):

src/PubSubHubbub/Subscriber.php
weierophinney: The only change here was the indentation. However, Psalm saw that, and raised an error.

But yes, at this point, the value is always a string here.

src/Reader/AbstractFeed.php
weierophinney: It's what the code was already doing; just documenting it properly here.

src/Reader/Entry/AbstractEntry.php
weierophinney: Same comment as above - the code was already doing this, but wasn't documented as doing it.

src/PubSubHubbub/Subscriber.php
boesing: Hm, usually, psalm does not care about indentation or the line where the error occured (as phpstan does).
It only stores "PossiblyInvalidCast appears 2 times in file XY" within the baseline. 🤔

src/Reader/Entry/Atom.php
weierophinney: I've added private functions for fetching each type of extension in each of the RSS and Atom entry classes now. Each attempts to retrieve the associated extension, tests that it is the correct type (throwing a `RuntimeException` if not), and returns it.

src/Reader/Extension/Atom/Feed.php
weierophinney: It **was** used, but Psalm wasn't seeing it. Regardless, I've rewritten the code, and the error is now gone.

src/Uri.php
weierophinney: But an `''` is a possible string value, and that value would be considered a non-absolute URI. Hence the check for `! empty()`.

src/Writer/ExtensionPluginManager.php
weierophinney: Parent is defined `array<array-key, string>`.

Ocramius (approved):
:ship:

PR #45, opened 2021-11-18
Return null when there is no feed link.

### LINKED ISSUE #44: feedLink() is broken, returns invalid data if not found
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  |  2.15

#### Summary

The `getFeedLink()` method is broken and returns invalid data.

#### Current behavior

`getFeedLink()` on an RSS feed falls back to the `Atom\Feed` extension, and calls it's `getFeedLink()` method.

`Atom\Feed::getLink()` first looks for a `<link rel="self" />` element to get a link.  If it finds nothing, it gets an empty string.  So far so good.

However, it then calls `absolutiseUrl()` on that value unconditionally, which when passed an empty string dutifully slaps the base domain of the feed onto the start of an empty string, giving you the domain of the site... Which is almost certainly not the `self` link of the feed, and may not even be related to the feed in the first place.  But that's what then gets returned.

#### How to reproduce

Feed (no pun intended) any RSS feed with no self link into it and call `getFeedLink()`.

#### Expected behavior

According to the method signatures, it should return null if there is no `self` link.

Crell:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Fixes #44.

At least I think so; the dev dependencies are fixed at PHP 8-incompatible versions so I can't actually run the code locally.  I think this is the right fix, but someone with an older PHP version available should check my work.

### DIFF
--- a/src/Reader/Extension/Atom/Feed.php
+++ b/src/Reader/Extension/Atom/Feed.php
@@ -330,7 +330,11 @@ class Feed extends Extension\AbstractFeed
 
         $link = $this->xpath->evaluate('string(' . $this->getXpathPrefix() . '/atom:link[@rel="self"]/@href)');
 
-        $link = $this->absolutiseUri($link);
+        if ($link) {
+            $link = $this->absolutiseUri($link);
+        } else {
+            $link = null;
+        }
 
         $this->data['feedlink'] = $link;
 


froschdesign:
@Crell 
> I think this is the right fix…

Thanks for this pull request but we need a test case for this change.

PR #52, opened 2022-07-18
Mention creation of Feeds in composer.json

heiglandreas:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This PR adds a mention that the package can also be used to create - and not only consume - feeds.

It also adds the "rss" and the "atom" keywords. Even though those two phrases are already part of the description it can not hurt to have them in the keywords as well.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,10 +1,12 @@
 {
     "name": "laminas/laminas-feed",
-    "description": "provides functionality for consuming RSS and Atom feeds",
+    "description": "provides functionality for creating and consuming RSS and Atom feeds",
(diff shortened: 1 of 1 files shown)


Ocramius:
Whoa, this library's CI is in shambles :scream:

heiglandreas:
Wanted to check that later today... But that was about my thought as well...

Ocramius:
LGTM: no need to rebase this.

PR #53, opened 2022-07-18
Fixes Markdown linting issues

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Fixes Markdown-lint problems with the docs

### DIFF
--- a/docs/book/find-feeds.md
+++ b/docs/book/find-feeds.md
@@ -8,6 +8,7 @@ feeds referenced by a web page with one method call:
 $feedLinks = Laminas\Feed\Reader\Reader::findFeedLinks('http://www.example.com/news.html');
 ```
 
+<!-- markdownlint-disable-next-line heading-increment -->
 > ### Finding feed links requires an HTTP client
 >
 > To find feed links, you will need to have an [HTTP client](http-clients.md)
--- a/docs/book/importing.md
+++ b/docs/book/importing.md
@@ -8,6 +8,7 @@ to consume it:
 $feed = Laminas\Feed\Reader\Reader::import('http://feeds.example.com/feedName');
 ```
 
+<!-- markdownlint-disable-next-line heading-increment -->
 > ### Importing requires an HTTP client
 >
 > To import a feed, you will need to have an [HTTP client](http-clients.md)
--- a/docs/book/intro.md
+++ b/docs/book/intro.md
@@ -18,6 +18,7 @@ In the example below, we demonstrate a simple use case of retrieving an RSS feed
 and saving relevant portions of the feed data to a simple PHP array, which could
 then be used for printing the data, storing to a database, etc.
 
+<!-- markdownlint-disable-next-line heading-increment -->
 > ### RSS optional properties
 >
 > Many *RSS* feeds have different channel and item properties available. The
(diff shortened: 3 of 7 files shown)


Ocramius:
A `composer install` to force it to generate the plugin map locally is needed

gsteel:
🙄 laminas-cache is too old for 8.0 without `--ignore-platform-reqs`. Update `.laminas-ci.json`? I can do that now…

gsteel:
Gonna have to come back to this later. Removing `laminas-cache` makes installation on 8.0 possible, but psalm complains about the unknown symbols.

_There are zero cache integration tests here_ so all the tests pass happily.

Ocramius:
IMO OK to expand baseline for now.

Ocramius:
Possibly worth declaring a conflict with newer cache releases, if the API is not compatible? :thinking:

gsteel:
> Possibly worth declaring a conflict with newer cache releases, if the API is not compatible?

Installing cache ^3 here appears to be compatible - it's only `Reader` that makes use of a cache (with a static property 🤢).

Docs are out of date, but it should be feasible to add 2||3 to dev deps

docs/book/intro.md
```diff
@@ -18,6 +18,7 @@ In the example below, we demonstrate a simple use case of retrieving an RSS feed
 and saving relevant portions of the feed data to a simple PHP array, which could
 then be used for printing the data, storing to a database, etc.
 
+<!-- markdownlint-disable-next-line heading-increment -->
 > ### RSS optional properties
```
froschdesign, line 22:
```suggestion
> NOTE: **RSS Optional Properties**
```

docs/book/importing.md
```diff
@@ -8,6 +8,7 @@ to consume it:
 $feed = Laminas\Feed\Reader\Reader::import('http://feeds.example.com/feedName');
 ```
 
+<!-- markdownlint-disable-next-line heading-increment -->
 > ### Importing requires an HTTP client
```
froschdesign, line 12:
```suggestion
> MISSING: Importing Requires an HTTP Client
```

docs/book/find-feeds.md
```diff
@@ -8,6 +8,7 @@ feeds referenced by a web page with one method call:
 $feedLinks = Laminas\Feed\Reader\Reader::findFeedLinks('http://www.example.com/news.html');
 ```
 
+<!-- markdownlint-disable-next-line heading-increment -->
 > ### Finding feed links requires an HTTP client
```
froschdesign, line 12:
```suggestion
> MISSING: Finding Feed Links Requires an HTTP Client
```

froschdesign (changes requested):
Admonitions / Call-Outs can be used here and no extra HTML comments / commands are needed.

(The admonitions are managed by the MkDocs plugin [markdown-callouts](https://oprypin.github.io/markdown-callouts/).)

Ocramius (approved):
Thanks @gsteel!

Ocramius:
@froschdesign argh, 1 minute race condition, sorry!  @_@

gsteel:
Gimme 5 and I'll send in a patch for @froschdesign's suggestions

PR #57, opened 2022-07-18
Update Laminas Coding Standard to 2.3, Remove Interop\Container, Drop Support for PHP 7.3

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

- Adds strict types
- Fixes calls to DomDocument#importNode to use `bool` for deep parameter
- Drops support for PHP 7.3
- Moves `laminas-servicemanager` into `require` at 3.14 minimum, making it a hard dependency and allowing the removal of `Interop\Container` and the ZF Bridge

### DIFF
--- a/src/Writer/Renderer/Entry/Rss.php
+++ b/src/Writer/Renderer/Entry/Rss.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Laminas\Feed\Writer\Renderer\Entry;
 
 use DateTime;
@@ -218,7 +220,7 @@ class Rss extends Renderer\AbstractRenderer implements Renderer\RendererInterfac
         }
         $enclosure = $this->dom->createElement('enclosure');
         $enclosure->setAttribute('type', $data['type']);
-        $enclosure->setAttribute('length', $data['length']);
+        $enclosure->setAttribute('length', (string) $data['length']);
         $enclosure->setAttribute('url', $data['uri']);
         $root->appendChild($enclosure);
     }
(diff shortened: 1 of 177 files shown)


src/Writer/Renderer/Entry/Rss.php
```diff
@@ -218,7 +220,7 @@ protected function _setEnclosure(DOMDocument $dom, DOMElement $root)
         }
         $enclosure = $this->dom->createElement('enclosure');
         $enclosure->setAttribute('type', $data['type']);
-        $enclosure->setAttribute('length', $data['length']);
+        $enclosure->setAttribute('length', (string) $data['length']);
```
Ocramius, line 223: Ok, this will likely not fly without a major release 😬

gsteel: L201:L220 assert the value is an integer - why would this be a BC break?

gsteel:
<del>This is going to need more work! Feel free to close it @Ocramius if you want to get 2.18 out the door</del>

src/Writer/Renderer/Entry/Rss.php
Ocramius: Ah, confused this with a test, sorry.

Ocramius (approved):
LGTM :+1:

Ocramius:
:ship: from my end.

I will probably regret it, but I checked this multiple times, and we don't seem to be calling userland functions from our own implementation.

In case of crashes, I will apply fixes first, and if that also fails, I will revert the strict types in another patch.

Ocramius:
Hah, it already [broke](https://github.com/laminas/laminas-feed/runs/7405125679?check_suite_focus=true) because `laminas/laminas-servicemanager` just got released (with `psr/container:^2` compat :D )

Awesome xD

Ocramius:
Related: https://github.com/laminas/laminas-servicemanager/pull/127#issuecomment-1188754196

Ocramius:
Fix provided in https://github.com/laminas/laminas-feed/pull/59

PR #62, opened 2022-08-08
Fix strict_types errors in DOMDocument::createTextNode()

### LINKED ISSUE #61: Error 500 when rendering a feed with some set*() methods
### BC Break Report

|    Q        |   A
|------------ | ------
| Version     | 2.18.1

#### Summary

Hello,

When I call setItunesEpisode(), i got this errror:

image

Since This [PR](https://github.com/laminas/laminas-feed/pull/57), `declare(strict_types=1);` was added and broke some pieces of code when argument of createTextNode() is not a string. Same behavior with setItunesSeason() (and maybe others)

#### Previous behavior

All good with v2.17, i can render an RSS feed with iTunes attributes.

#### Current behavior

Error 500, because of a wrong parameter type.

#### How to reproduce

Just Write a simple feed :

        $feed = new \Laminas\Feed\Writer\Feed ();

        $feed->setTitle('Serie Name')
            ->setDescription('Description')
            ->setLink('https//www.example.com');

        $item = $feed->createEntry();
        $item->setTitle('Episode Title')
            ->setItunesEpisode(3);

        $feed->addEntry($item);

        $feed->export('rss');

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Modifies all calls to `DOMDocument::createTextNode()` to ensure parameters are cast to a string where the type is currently ambiguous.

Should close #61

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -2223,9 +2223,8 @@
     </MixedMethodCall>
   </file>
   <file src="src/Writer/Extension/DublinCore/Renderer/Entry.php">
-    <MixedArgument occurrences="2">
+    <MixedArgument occurrences="1">
       <code>$data</code>
-      <code>$data['name']</code>
     </MixedArgument>
     <MixedAssignment occurrences="2">
       <code>$authors</code>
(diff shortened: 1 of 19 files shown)


Ocramius (approved):
👍 thanks @gsteel!

PR #66, opened 2022-11-15
Initialize type as an empty string instead of null

LuomaJuha:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
Passing null into strtolower is deprecated in php 8.1 version. Initializing type as an empty string prevents this from happening.

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,9 +1,8 @@
 name: docs-build
 
 on:
-  push:
-    branches:
-      - master
+  release:
+    types: [published]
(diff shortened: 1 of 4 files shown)


Ocramius (changes requested):
There should be a test highlighting this change requirement, IMO.

Ocramius:
This patch targeted a very ancient `2.12.x`: I re-targeted `2.19.x`, since this is a bugfix, but a test and a rebase are still needed.

PR #74, opened 2023-06-06
Update composer.json - php 8.2 dependancy update

RicLeP:
### Description

Update `laminas/laminas-servicemanager` in composer.json to support php8.2


Ocramius:
This change has no effect: closing.

Probably a misunderstanding by the PR author in how `^` works in composer.

PR #81, opened 2023-12-09
Move header value normalisation into Response

thgs:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | yes
| QA            | yes/no

### Description

When not using laminas-http as the HTTP client the implementation of the `HeaderAwareClientInterface` one may have to reproduce the logic to normalise headers from multi value to string as it is done in [here](https://github.com/laminas/laminas-feed/blob/2.23.x/src/Reader/Http/LaminasHttpClientDecorator.php#L106). 

I am partial about whether what I am doing here is right (or desired) or wrong (will have problems), so it is more of a question, in the form of a PR for clarity.

The benefit of this change is that when implementing the interfaces needed to inject a third party client, one does not need to deal with the case of multi value headers and can reuse the Response class provided (`Laminas\Feed\Reader\Http\Response`), therefore only one implementation is needed implementing `HeaderAwareClientInterface`. 

I am relatively new to the codebase so let me know if I have missed some profound/foundational concern here. I can see a slight violation of single responsibility principle here.

On the other hand, it is a (micro?) optimisation as the loop through the headers only happens once. 

Lastly, given this is desired, maybe a unit test case should be added as well as an addition in the docs explaining the above "convenience". Ofc I am willing to do these and any other changes requested but I thought it would be nice if first it is acknowledged that the direction is desired.

### DIFF
--- a/src/Reader/Http/LaminasHttpClientDecorator.php
+++ b/src/Reader/Http/LaminasHttpClientDecorator.php
@@ -9,7 +9,6 @@ use Laminas\Http\Client as LaminasHttpClient;
 use Laminas\Http\Headers;
 
 use function gettype;
-use function implode;
 use function is_array;
 use function is_numeric;
 use function is_object;
@@ -47,10 +46,12 @@ class LaminasHttpClientDecorator implements HeaderAwareClientInterface
         }
         $response = $this->client->send();
 
+        $headers = $response->getHeaders()->toArray();
+
         return new Response(
             $response->getStatusCode(),
             $response->getBody(),
-            $this->prepareResponseHeaders($response->getHeaders())
+            $headers
         );
     }
 
@@ -93,22 +94,5 @@ class LaminasHttpClientDecorator implements HeaderAwareClientInterface
                 $headers->addHeaderLine($name, $value);
             }
         }
-    }
-
-    /**
-     * Normalize headers to use with HeaderAwareResponseInterface.
-     *
-     * Ensures multi-value headers are represented as a single string, via
-     * comma concatenation.
-     *
-     * @return array
-     */
-    private function prepareResponseHeaders(Headers $headers)
-    {
-        $normalized = [];
-        foreach ($headers->toArray() as $name => $value) {
-            $normalized[$name] = is_array($value) ? implode(', ', $value) : $value;
-        }
-        return $normalized;
     }
 }
--- a/src/Reader/Http/Response.php
+++ b/src/Reader/Http/Response.php
@@ -144,13 +144,26 @@ class Response implements HeaderAwareResponseInterface
                 ));
             }
 
-            if (! is_string($value) && ! is_numeric($value)) {
+            if (! is_string($value) && ! is_numeric($value) && ! is_array($value)) {
                 throw new Exception\InvalidArgumentException(sprintf(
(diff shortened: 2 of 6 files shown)


Ocramius:
> When not using laminas-http as the HTTP client

Only concern is whether we are breaking any pre-existing behaviour.

I suggest:

 * Documenting the array types (inputs/outputs) that touch the changed code
 * adding tests that prevents regressions on multi-valued headers with laminas-http and without

src/Reader/Http/LaminasHttpClientDecorator.php
```diff
@@ -47,10 +46,12 @@ public function get($uri, array $headers = [])
         }
         $response = $this->client->send();
 
+        $headers = $response->getHeaders()->toArray() ?? [];
```
Ocramius: Im theory this is always an array?

src/Reader/Http/Response.php
```diff
@@ -155,14 +157,16 @@ private function validateHeaders(array $headers)
 
     /**
      * Normalize header names to lowercase.
+     * Also ensures multi-value headers are represented as a single string, via
+     * comma concatenation.
      *
      * @return array
```
Ocramius, line 176: Let's document this to be an `array<string>` then? Previously, it was an `array<string|array<string>>`, so this change looks like a breaking change 🤔

Ocramius (changes requested):

src/Reader/Http/LaminasHttpClientDecorator.php
thgs: Eh yes, forgot the `?->`, will fix it. `getHeaders` is returning `null|Headers`

src/Reader/Http/Response.php
thgs: Its the opposite I think. Before the `normalizeHeaders` was expecting `array<string, string>` and now it can handle `array<string, string|array<string>>`, as the logic that was in [LaminasHttpClientDecorator prepareResponseHeaders](https://github.com/laminas/laminas-feed/blob/2.23.x/src/Reader/Http/LaminasHttpClientDecorator.php#L106) now is included in Response `normalizeHeaders` (see `validateHeaders` in `Response` is not accepting multi value (L147 in my branch).

So I thought its not a breaking change as now it can handle what it was handling before and in addition multi-value headers?

thgs:
>  When not using laminas-http as the HTTP client
> 
> Only concern is whether we are breaking any pre-existing behaviour.
> 
> I suggest:
> 
>     * Documenting the array types (inputs/outputs) that touch the changed code
> 
>     * adding tests that prevents regressions on multi-valued headers with laminas-http and without

Yes, will check more myself for BC breaks and do the suggestions soon.

src/Reader/Http/LaminasHttpClientDecorator.php
thgs: Oh I see what you mean now. `AbstractMessage` (Laminas\Http) handles null. IDE was showing it `null|Headers` for whatever reason. `null` is handled on getHeaders.

thgs:
@Ocramius 

Happy new year.

I think I am going to back off from this for now. The benefit is very little and it seems more risky the more I think about it.

Thanks for review!

PR #88, opened 2024-09-16
Fix type-safety for `ExtensionPluginManager`

RobLoach:
This fixes a few type definitions that were missing when defining instances within `ExtensionPluginManager`, resulting in a complete white-screen when `error_reporting(E_ALL)` is enabled.

For instance, [`PluginManagerInterface::validate`](https://github.com/laminas/laminas-servicemanager/blob/4.3.x/src/PluginManagerInterface.php#L27) expects to return `void`.

This change fixes the following...

>Got error 'PHP message: PHP Fatal error:  Declaration of Laminas\\Feed\\Reader\\ExtensionPluginManager::validate(mixed $instance) must be compatible with Laminas\\ServiceManager\\PluginManagerInterface::validate(mixed $instance): void in /var/www/html/vendor/laminas/laminas-feed/src/Reader/ExtensionPluginManager.php on line 203'

|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

- Are you fixing a bug or providing a failing unit test to demonstrate a bug? Yes, this fixes type safety issues in the inherited class.
  - How do you reproduce it? Use `error_reporting(E_ALL)`, and inherit from `laminas-service` 4.3
  - What did you expect to happen? For there to not be any type errors.
  - What actually happened? It reported the above error
  - TARGET THE CURRENT RELEASE BRANCH

- Are you adding documentation? No.
  - TARGET THE CURRENT RELEASE BRANCH

- Are you providing a QA improvement (additional tests, CS fixes, etc.) that
  does not change behavior? No, there is no behavior change.
  - Explain why the changes are necessary. The change is required for type-safety across the PHP classes.
  - TARGET THE NEXT MINOR BRANCH

- Are you fixing a BC Break? Unsure.
  - How do you reproduce it? Use `error_reporting(E_ALL)`
  - What was the previous behavior? Things work without the error reporting.
  - What is the current behavior? Things break when you're using error reporting.
  - TARGET THE CURRENT RELEASE BRANCH

- Are you adding something the library currently does not support? No.

- Are you refactoring code? No.

### DIFF
--- a/src/Reader/ExtensionPluginManager.php
+++ b/src/Reader/ExtensionPluginManager.php
@@ -33,7 +33,7 @@ class ExtensionPluginManager extends AbstractPluginManager implements ExtensionM
      *
      * @inheritDoc
      */
-    protected $aliases = [
+    protected array $aliases = [
         'atomentry'               => Extension\Atom\Entry::class,
         'atomEntry'               => Extension\Atom\Entry::class,
         'AtomEntry'               => Extension\Atom\Entry::class,
@@ -143,7 +143,7 @@ class ExtensionPluginManager extends AbstractPluginManager implements ExtensionM
      *
      * @inheritDoc
      */
-    protected $factories = [
+    protected array $factories = [
         Extension\Atom\Entry::class              => InvokableFactory::class,
         Extension\Atom\Feed::class               => InvokableFactory::class,
         Extension\Content\Entry::class           => InvokableFactory::class,
@@ -197,10 +197,10 @@ class ExtensionPluginManager extends AbstractPluginManager implements ExtensionM
      *
      * @var bool
      */
-    protected $sharedByDefault = false;
+    protected bool $sharedByDefault = false;
 
     /** @inheritDoc */
-    public function validate(mixed $instance)
+    public function validate(mixed $instance) : void
     {
         if (
             $instance instanceof AbstractEntry


RobLoach:
The `NonInvariantPropertyType` errors should be present in the parent classes as well.

Xerkus:
Adding types to properties and return types would be a BC break. 

This package is not compatible with service manager v4. It is an optional dependency so no composer constraints. Conflict with v4 should be declared that can be resolved only in the next major.

RobLoach:
Thanks for the explanation, Xerkus. Are there plans to make it compatible with v4? Anything I can do to help make it happen?

Xerkus:
A new major version is needed but it can't really be done in multiple small steps. Before it can be tagged a major QA overhaul to modernize code to the current PHP, improve types and psalm coverage, resolve any other breaking issues and so on.

If you are willing to tackle those tasks we can start work on a new major release.

froschdesign:
@RobLoach 
The `ExtensionPluginManager` is optional, [you can use the `StandaloneExtensionManager` instead](https://docs.laminas.dev/laminas-feed/reader/#extending-feed-and-entry-apis).

And an update is planned, but first all the foundational components [need to be updated](https://github.com/laminas/laminas-servicemanager/issues/216).

PR #97, opened 2025-05-12
Bugfix PodcastIndex: Set funding tag, not locked

nb823:
- Fixed Feed\Writer\Renderer::setPodcastIndexFunding() to generate tag <podcast:funding> instead of tag <podcast:locked>
- Created tests to assert the correct tags are created by the Feed\Writer\Renderer methods

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /laminas-mkdoc-theme/
 /phpunit.xml
 /vendor/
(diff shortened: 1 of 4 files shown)


nb823:
Changes were made to branch 2.24.x not 2.23.x.
I will drop the changes and redo them on the correct branch.

Issue #96, opened 2025-05-06
Method setPodcastIndexFunding() creates wrong tag


nb823:
|    Q        |   A
|------------ | ------
| Version(s)  | 2.23

#### Summary

The method setPodcastIndexFunding() (that is provided by the PodcastIndex extension of this package `Laminas\Feed\Writer\Extension\PodcastIndex\Feed`) creates an unexpected tag `<podcast:locked>` instead of the expected `<podcast:funding>`

#### How to reproduce

```
$feedWriter = new \Laminas\Feed\Writer\Feed();
$value = [
         'title' => 'Support the show!',
         'url' => 'https://www.support-my-show-yay.com',
];
$feedWriter->setPodcastIndexFunding($value);
```

#### Current behavior

Generates this tag in an rss feed:
`<podcast:locked url="https://www.support-my-show-yay.com">Support the show!</podcast:locked>`

#### Expected behavior

Should generate this tag in an rss feed:
`<podcast:funding url="https://www.support-my-show-yay.com">Support the show!</podcast:funding>`

froschdesign:
@nb823 
Can you create a pull request as bugfix? Please use the branch 2.23.x.
Thanks in advance! 👍🏻 

https://github.com/laminas/laminas-feed/blob/23807e692b3174750b426143bd93572b71b6739a/src/Writer/Extension/PodcastIndex/Renderer/Feed.php#L70-L78

nb823:
@froschdesign 
Please see my pull request: https://github.com/laminas/laminas-feed/pull/97
This is my first time contributing to a GitHub project by forking, so I hope I am doing ok :)

PR #98, opened 2025-05-12
Fix: PodcastIndex feed extension and tests

nb823:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

Fix for Issue: https://github.com/laminas/laminas-feed/issues/96

- Made a small fix on the PodcastIndex extension renderer
- Added two tests to check on the feed podcastindex tags generated

### DIFF
--- a/test/Writer/Renderer/Extension/PodcastIndex/FeedTest.php
+++ b/test/Writer/Renderer/Extension/PodcastIndex/FeedTest.php
@@ -5,12 +5,12 @@ declare(strict_types=1);
 namespace LaminasTest\Feed\Writer\Renderer\Extension\PodcastIndex;
 
 use Laminas\Feed\Writer;
-use Laminas\Feed\Reader;
-use PHPUnit\Framework\TestCase;
 use Laminas\Feed\Writer\Renderer;
+use PHPUnit\Framework\TestCase;
+
 class FeedTest extends TestCase
 {
-    protected $validWriter;
+    protected Writer\Feed $validWriter;
 
     protected function setUp(): void
     {
@@ -35,12 +35,10 @@ class FeedTest extends TestCase
             'owner' => 'john.doe@example.com',
         ];
         $this->validWriter->setPodcastIndexLocked($locked);
-        
-        $rssFeed = new Renderer\Feed\Rss($this->validWriter);
-        $rssFeed->render();
-        $feed = Reader\Reader::importString($rssFeed->saveXml());
-        $xml = $feed->getDomDocument()->saveXml();
 
+        $rssFeed = new Renderer\Feed\Rss($this->validWriter);
+        /** @var string $xml */
+        $xml = $rssFeed->render()->saveXml();
         $this->assertStringContainsString('<podcast:locked', $xml);
     }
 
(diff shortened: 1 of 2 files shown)


test/Writer/Renderer/Extension/PodcastIndex/FeedTest.php
```diff
@@ -0,0 +1,59 @@
...
+    public function testRendersLockedTag(): void
+    {
+        $locked = [
+            'value' => 'yes',
+            'owner' => 'john.doe@example.com',
+        ];
+        $this->validWriter->setPodcastIndexLocked($locked);
+
+        $rssFeed = new Renderer\Feed\Rss($this->validWriter);
+        /** @var string $xml */
```
gsteel: If you're worried this might not be a string, add an assertion after the assignment, otherwise just drop it to stop Psalm complaining.

test/Writer/Renderer/Extension/PodcastIndex/FeedTest.php
```diff
@@ -0,0 +1,59 @@
...
+    public function testRendersFundingTag(): void
+    {
+        $funding = [
+            'title' => 'Support the show!',
+            'url' => 'http://example.com/donate',
+        ];
+        $this->validWriter->setPodcastIndexFunding($funding);
+
+        $rssFeed = new Renderer\Feed\Rss($this->validWriter);
+        /** @var string $xml */
```
gsteel: Same here with `@var` switch to an assertion, or simply remove it.

test/Writer/Renderer/Extension/PodcastIndex/FeedTest.php
```diff
@@ -0,0 +1,59 @@
...
+        $xml = $rssFeed->render()->saveXml();
+        $this->assertStringContainsString('<podcast:locked', $xml);
+    }
+
+    public function testRendersFundingTag(): void
+    {
+        $funding = [
+            'title' => 'Support the show!',
+            'url' => 'http://example.com/donate',
```
gsteel:
```suggestion
            'url'   => 'http://example.com/donate',
```

gsteel (changes requested):
LGTM @nb823 - Just a couple of tweaks to satisfy CI and this can be merged & released

nb823:
@gsteel Unfortunately, my local CodeSniffer and Psalm plugins stumble upon different lines of code, so that I only see the above problems after pushing the new commits. The lines you mentioned above were already just a reaction on the previous check fails I got here. Now, after another update, there is still simple formatting issues open that I don't know how to solve, since I cannot see what exact lines are failing.

gsteel:
@nb823 Assuming you have run `composer install` locally, from the project root `vendor/bin/phpcs`, `vendor/bin/phpcbf` and `vendor/bin/psalm` should all work "out of the box", (also assuming you have a supported version of PHP available locally)

gsteel:
If Psalm is causing problems - Also try `vendor/bin/psalm --no-cache --threads=1` to verify SA passes

gsteel (approved):
Thank you @nb823! Much appreciated

nb823:
> from the project root `vendor/bin/phpcs`, `vendor/bin/phpcbf` and `vendor/bin/psalm` should all work "out of the box"

Thank you for the hint to run those scripts directly @gsteel 
Now it passes all checks.

PR #100, opened 2025-06-18
Podcastindex extensions for feed

### LINKED ISSUE #95: Expand Support for PodcastIndex
|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

The package already offers a bit of support for PodcastIndex via `Laminas\Feed\Writer\Extension\PodcastIndex\Feed` and `Laminas\Feed\Writer\Extension\PodcastIndex\Entry`. But there are only a few methods offered, for only a few PodcastIndex namespaces.

The Laminas docs list the already supported namespaces that you can also find in the above feed writer extension classes of the package: https://docs.laminas.dev/laminas-feed/extensions/podcast-index/

**The supported namespaces are:**

-  Feed: Locked
-  Feed: Funding
-  Entry: Transcript
-  Entry: Chapters
-  Entry: Soundbite

And this is an overview of all the PodcastIndex namespaces that would be great to be completely supported by this package: https://podcasting2.org/podcast-namespace

**The missing namespaces:**

-  Feed: Podroll
-  Feed: Person
-  Feed: Location
-  (Feed: License)
-  Feed: Trailer
-  Feed: Guid
-  Feed: Value
-  Feed: Value Recipient
-  Feed: Medium
-  Feed: Images
-  Feed: Live Item
-  Feed: Content Link
-  Feed: Block
-  Feed: Txt
-  Feed: Remote Item
-  Feed: Update Frequency
-  Feed: Podping
-  Feed: Value Time Split
-  Entry: Person
-  Entry: Location
-  Entry: Season
-  Entry: Episode
-  Entry: License
-  Entry: Alternate Inclosure
-  Entry: Source
-  Entry: Integrity
-  Entry: Value
-  Entry: Value Recipient
-  Entry: Images
-  Entry: Social Interact
-  Entry: Txt

nb823:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

1.  Extended PodcastIndex support for additional namespaces:

- Feed License
- Feed Location
- Feed Images
- Feed UpdateFrequency
- Feed Person

2. Added test for the above namespaces
3. Added documentation for the above namespaces
4. Introduced new naming convention for PodcastIndex get methods (example: getPodcastIndexLicense() to be distinguished from getLicense())

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /laminas-mkdoc-theme/
 /phpunit.xml
 /vendor/
+/.idea
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@
 /laminas-mkdoc-theme/
 /phpunit.xml
 /vendor/
-/.idea
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -3107,6 +3107,12 @@
       <code>setEncoding</code>
       <code>setPodcastIndexFunding</code>
       <code>setPodcastIndexLocked</code>
+      <code>setPodcastIndexLicense</code>
+      <code>setPodcastIndexLocation</code>
+      <code>setPodcastIndexImages</code>
+      <code>setPodcastIndexUpdateFrequency</code>
+      <code>addPodcastIndexPerson</code>
+      <code>setPodcastIndexPersons</code>
     </PossiblyUnusedMethod>
     <PossiblyUnusedParam>
       <code>$params</code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.18.0@b113f3ed0259fd6e212d87c3df80eec95a6abf19">
+<files psalm-version="5.19.0@06b71be009a6bd6d81b9811855d6629b9fe90e1b">
   <file src="src/PubSubHubbub/AbstractCallback.php">
     <DocblockTypeContradiction>
       <code><![CDATA[$this->httpResponse === null]]></code>
@@ -3096,18 +3096,34 @@
     </PossiblyUnusedProperty>
   </file>
   <file src="src/Writer/Extension/PodcastIndex/Feed.php">
+    <DocblockTypeContradiction>
+      <code><![CDATA[! is_string($value['url'])]]></code>
+      <code><![CDATA[is_string($value['description'])]]></code>
+      <code><![CDATA[is_string($value['description'])]]></code>
+      <code><![CDATA[is_string($value['identifier'])]]></code>
+      <code><![CDATA[is_string($value['name'])]]></code>
+      <code><![CDATA[is_string($value['srcset'])]]></code>
+      <code><![CDATA[isset($value['complete']) && ! is_bool($value['complete'])]]></code>
+      <code><![CDATA[isset($value['dtstart']) && ! is_string($value['dtstart'])]]></code>
+      <code><![CDATA[isset($value['geo']) && ! is_string($value['geo'])]]></code>
+      <code><![CDATA[isset($value['osm']) && ! is_string($value['osm'])]]></code>
+      <code><![CDATA[isset($value['rrule']) && ! is_string($value['rrule'])]]></code>
+    </DocblockTypeContradiction>
+    <MixedArgumentTypeCoercion>
+      <code>$value</code>
+    </MixedArgumentTypeCoercion>
     <PossiblyUnusedMethod>
       <code>__call</code>
       <code>__construct</code>
       <code>getEncoding</code>
       <code>setEncoding</code>
       <code>setPodcastIndexFunding</code>
-      <code>setPodcastIndexLocked</code>
+      <code>setPodcastIndexImages</code>
       <code>setPodcastIndexLicense</code>
       <code>setPodcastIndexLocation</code>
-      <code>setPodcastIndexImages</code>
-      <code>setPodcastIndexUpdateFrequency</code>
+      <code>setPodcastIndexLocked</code>
       <code>setPodcastIndexPersons</code>
+      <code>setPodcastIndexUpdateFrequency</code>
     </PossiblyUnusedMethod>
     <PossiblyUnusedParam>
       <code>$params</code>
@@ -3132,12 +3148,12 @@
   <file src="src/Writer/Extension/PodcastIndex/Renderer/Feed.php">
     <MixedMethodCall>
       <code>getPodcastIndexFunding</code>
-      <code>getPodcastIndexLocked</code>
+      <code>getPodcastIndexImages</code>
       <code>getPodcastIndexLicense</code>
       <code>getPodcastIndexLocation</code>
-      <code>getPodcastIndexImages</code>
-      <code>getPodcastIndexUpdateFrequency</code>
+      <code>getPodcastIndexLocked</code>
       <code>getPodcastIndexPersons</code>
+      <code>getPodcastIndexUpdateFrequency</code>
     </MixedMethodCall>
     <RedundantCastGivenDocblockType>
       <code><![CDATA[(string) $funding['title']]]></code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -3122,7 +3122,7 @@
       <code>setPodcastIndexLicense</code>
       <code>setPodcastIndexLocation</code>
       <code>setPodcastIndexLocked</code>
-      <code>setPodcastIndexPersons</code>
+      <code>setPodcastIndexPeople</code>
       <code>setPodcastIndexUpdateFrequency</code>
     </PossiblyUnusedMethod>
     <PossiblyUnusedParam>
@@ -3152,7 +3152,7 @@
       <code>getPodcastIndexLicense</code>
       <code>getPodcastIndexLocation</code>
       <code>getPodcastIndexLocked</code>
-      <code>getPodcastIndexPersons</code>
+      <code>getPodcastIndexPeople</code>
       <code>getPodcastIndexUpdateFrequency</code>
     </MixedMethodCall>
     <RedundantCastGivenDocblockType>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -3106,14 +3106,11 @@
       <code><![CDATA[isset($value['complete']) && ! is_bool($value['complete'])]]></code>
       <code><![CDATA[isset($value['dtstart']) && ! is_string($value['dtstart'])]]></code>
       <code><![CDATA[isset($value['geo']) && ! is_string($value['geo'])]]></code>
-      <code><![CDATA[isset($value['osm']) && ! is_string($value['osm'])]]></code>
-      <code><![CDATA[isset($value['rrule']) && ! is_string($value['rrule'])]]></code>
-      <code><![CDATA[isset($value['role']) && ! is_string($value['role'])]]></code>
       <code><![CDATA[isset($value['group']) && ! is_string($value['group'])]]></code>
+      <code><![CDATA[isset($value['osm']) && ! is_string($value['osm'])]]></code>
+      <code><![CDATA[isset($value['role']) && ! is_string($value['role'])]]></code>
+      <code><![CDATA[isset($value['rrule']) && ! is_string($value['rrule'])]]></code>
     </DocblockTypeContradiction>
-    <MixedArgumentTypeCoercion>
-      <code>$value</code>
-    </MixedArgumentTypeCoercion>
     <PossiblyUnusedMethod>
       <code>__call</code>
       <code>__construct</code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -3104,7 +3104,6 @@
       <code><![CDATA[is_string($value['name'])]]></code>
       <code><![CDATA[is_string($value['srcset'])]]></code>
       <code><![CDATA[isset($value['complete']) && ! is_bool($value['complete'])]]></code>
-      <code><![CDATA[isset($value['dtstart']) && ! is_string($value['dtstart'])]]></code>
       <code><![CDATA[isset($value['geo']) && ! is_string($value['geo'])]]></code>
       <code><![CDATA[isset($value['group']) && ! is_string($value['group'])]]></code>
       <code><![CDATA[isset($value['osm']) && ! is_string($value['osm'])]]></code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -3144,15 +3144,6 @@
     </RedundantCastGivenDocblockType>
   </file>
   <file src="src/Writer/Extension/PodcastIndex/Renderer/Feed.php">
-    <MixedMethodCall>
-      <code>getPodcastIndexFunding</code>
-      <code>getPodcastIndexImages</code>
-      <code>getPodcastIndexLicense</code>
-      <code>getPodcastIndexLocation</code>
-      <code>getPodcastIndexLocked</code>
-      <code>getPodcastIndexPeople</code>
-      <code>getPodcastIndexUpdateFrequency</code>
-    </MixedMethodCall>
     <RedundantCastGivenDocblockType>
       <code><![CDATA[(string) $funding['title']]]></code>
       <code><![CDATA[(string) $locked['value']]]></code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1582,9 +1582,6 @@
     </MixedReturnStatement>
   </file>
   <file src="src/Reader/Extension/PodcastIndex/Feed.php">
-    <MismatchingDocblockReturnType>
-      <code>null|object{url: string, title: string}</code>
-    </MismatchingDocblockReturnType>
     <MixedAssignment>
       <code>$locked</code>
       <code>$owner</code>
@@ -1592,7 +1589,6 @@
     <MixedInferredReturnType>
       <code>?string</code>
       <code>bool</code>
-      <code>null|object{url: string, title: string}</code>
     </MixedInferredReturnType>
     <MixedReturnStatement>
       <code><![CDATA[$this->data['locked']]]></code>
@@ -3111,7 +3107,6 @@
       <code>setPodcastIndexLocation</code>
       <code>setPodcastIndexImages</code>
       <code>setPodcastIndexUpdateFrequency</code>
-      <code>addPodcastIndexPerson</code>
       <code>setPodcastIndexPersons</code>
     </PossiblyUnusedMethod>
     <PossiblyUnusedParam>
@@ -3138,6 +3133,11 @@
     <MixedMethodCall>
       <code>getPodcastIndexFunding</code>
       <code>getPodcastIndexLocked</code>
+      <code>getPodcastIndexLicense</code>
+      <code>getPodcastIndexLocation</code>
+      <code>getPodcastIndexImages</code>
+      <code>getPodcastIndexUpdateFrequency</code>
+      <code>getPodcastIndexPersons</code>
     </MixedMethodCall>
     <RedundantCastGivenDocblockType>
       <code><![CDATA[(string) $funding['title']]]></code>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -3108,6 +3108,8 @@
       <code><![CDATA[isset($value['geo']) && ! is_string($value['geo'])]]></code>
       <code><![CDATA[isset($value['osm']) && ! is_string($value['osm'])]]></code>
       <code><![CDATA[isset($value['rrule']) && ! is_string($value['rrule'])]]></code>
+      <code><![CDATA[isset($value['role']) && ! is_string($value['role'])]]></code>
+      <code><![CDATA[isset($value['group']) && ! is_string($value['group'])]]></code>
     </DocblockTypeContradiction>
     <MixedArgumentTypeCoercion>
       <code>$value</code>
--- a/src/Reader/Extension/PodcastIndex/Feed.php
+++ b/src/Reader/Extension/PodcastIndex/Feed.php
@@ -107,6 +107,33 @@ class Feed extends Extension\AbstractFeed
     }
 
     /**
+     * Get the podcast location
+     *
+     * @psalm-return null|object{text: string, geo: string, osm: string}
+     */
+    public function getLocation(): ?stdClass
+    {
+        if (array_key_exists('location', $this->data)) {
+            return $this->data['location'];
+        }
+
+        $location = null;
+
+        $nodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:location');
+
+        if ($nodeList->length > 0) {
+            $location              = new stdClass();
+            $location->description = $nodeList->item(0)->nodeValue;
+            $location->geo         = $nodeList->item(0)->getAttribute('geo');
+            $location->osm         = $nodeList->item(0)->getAttribute('osm');
+        }
+
+        $this->data['location'] = $location;
+
+        return $this->data['location'];
+    }
+
+    /**
      * Register PodcastIndex namespace
      */
     protected function registerNamespaces(): void
--- a/src/Reader/Extension/PodcastIndex/Feed.php
+++ b/src/Reader/Extension/PodcastIndex/Feed.php
@@ -134,6 +134,31 @@ class Feed extends Extension\AbstractFeed
     }
 
     /**
+     * Get the podcast images
+     *
+     * @psalm-return null|object{scrset: string}
+     */
+    public function getImages(): ?stdClass
+    {
+        if (array_key_exists('images', $this->data)) {
+            return $this->data['images'];
+        }
+
+        $images = null;
+
+        $nodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:images');
+
+        if ($nodeList->length > 0) {
+            $images         = new stdClass();
+            $images->srcset = $nodeList->item(0)->getAttribute('srcset');
+        }
+
+        $this->data['images'] = $images;
+
+        return $this->data['images'];
+    }
+
+    /**
      * Register PodcastIndex namespace
      */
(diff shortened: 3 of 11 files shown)


src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -80,6 +78,135 @@ public function getFunding(): ?stdClass
...
+        }
+
+        $location = null;
+
+        $nodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:location');
+
+        if ($nodeList->length > 0) {
+            $location              = new stdClass();
+            $location->description = $nodeList->item(0)->nodeValue;
+            $location->geo         = $nodeList->item(0)->getAttribute('geo');
```
gsteel: You'll have to extract `$nodeList->item(0)` into a variable and assert it is an instance of `DomElement` before calling `getAttribute` on it.

.gitignore
```diff
@@ -8,3 +8,4 @@
 /laminas-mkdoc-theme/
 /phpunit.xml
 /vendor/
+/.idea
```
gsteel: This should be added to your personal `~/.gitignore_global`

froschdesign: See also in the documentation of Git: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

nb823:
Thanks @gsteel and @froschdesign for your hints!
All issues have been resolved now.

In my last commit, I also introduced a naming convention for the PodcastIndex get methods to prevent conflicts with other already existing methods (like getLicense() for example). PodcastIndex methods that are already in use are still callable by their old names.

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +110,189 @@ public function setPodcastIndexFunding(array $value): Feed
         return $this;
     }
 
+    /**
+     * Set feed license
+     *
+     * @param array $value [identifier: string, url: string]
```
gsteel: Could you update setters that accept an array with the expected shape please:

```suggestion
     * @param array{identifier: string, url: string} $value
```

You'll get a load of `RedundantCondition` issues from Psalm regarding the exceptions which can be baselined.

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +110,189 @@ public function setPodcastIndexFunding(array $value): Feed
...
+            throw new Writer\Exception\InvalidArgumentException(
+                'invalid parameter: "license" must be an array containing the keys "identifier" (node value) and "url"'
+            );
+        }
+        if (! is_string($value['identifier'])) {
+            throw new Writer\Exception\InvalidArgumentException(
+                'invalid parameter: "identifier" of "license" must be of type string.'
+            );
+        }
+        if (! is_string($value['url']) || ! filter_var($value['url'], FILTER_VALIDATE_URL)) {
```
gsteel, line 147: Please can you make sure that `ext-filter: *` is `require`'d in `composer.json`

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +110,189 @@ public function setPodcastIndexFunding(array $value): Feed
...
+            );
+        }
+        $this->data['license'] = $value;
+        return $this;
+    }
+
+    /**
+     * Set feed location
+     *
+     * @param array $value [description: string, geo: string|null, osm: string|null]
```
gsteel: Same thing with the array shape here please

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,188 @@ public function getLockOwner(): ?string
...
+            $funding->title = $item->nodeValue;
         }
 
         $this->data['funding'] = $funding;
 
         return $this->data['funding'];
     }
 
+    /**
+     * Get the podcast license
```
gsteel, line 121: Can you update all these getters to include the object shape please:

```suggestion
     * Get the podcast license
     * @return object{identifier: string, url: string}
```

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,188 @@ public function getLockOwner(): ?string
...
+     */
+    public function getPodcastIndexLicense(): ?stdClass
```
gsteel: For all new methods, can we return `object` instead of `stdClass`

```suggestion
    public function getPodcastIndexLicense(): object|null
```

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +110,189 @@ public function setPodcastIndexFunding(array $value): Feed
...
+     * @return $this
+     * @throws Writer\Exception\InvalidArgumentException
+     */
+    public function setPodcastIndexLocation(array $value)
```
gsteel: Please add the return type to all newly added methods

```suggestion
    public function setPodcastIndexLocation(array $value): self
```

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +110,189 @@ public function setPodcastIndexFunding(array $value): Feed
...
+            );
+        }
+        $this->data['images'] = $value;
+        return $this;
+    }
+
+    /**
+     * Set feed update frequency
+     *
+     * @param array $value [description: string, complete: bool|null, dtstart: string|null, rrule: string|null]
```
gsteel: Optional keys in array shapes are `array{optional?: type}` if it's OK for the key to contain null, it'd be `array{optional?: type|null}`

```suggestion
     * @param array{description: string, complete?: bool, dtstart?: string, rrule?: string} $value
```

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +110,189 @@ public function setPodcastIndexFunding(array $value): Feed
...
+
+    /**
+     * Set a new array of persons.
+     * If no argument is passed, it will just remove all existing persons.
+     *
+     * @param null|array<array> $values
+     * @return $this
+     * @throws Writer\Exception\InvalidArgumentException
+     */
+    public function setPodcastIndexPersons(?array $values = null)
```
gsteel: Why not:
```suggestion
    public function setPodcastIndexPersons(array $values = []): self
```

Also, why "Persons" as opposed to "People" - does this follow the naming in the podcast index spec?

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +110,189 @@ public function setPodcastIndexFunding(array $value): Feed
...
+    {
+        // delete existing persons before setting new ones
+        $this->data['persons'] = null;
```
gsteel: With `$values` always an array

```suggestion
        $this->data['persons'] = [];
```

And you can drop the subsequent conditional

gsteel (changes requested):
Thanks for your work here @nb823

A few adjustments required :)

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,198 @@ public function getLockOwner(): ?string
...
+            return $object;
         }
 
         $funding = null;
 
         $nodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:funding');
 
         if ($nodeList->length > 0) {
+            /** @var DOMElement $item */
+            $item           = $nodeList->item(0);
```
gsteel:
```suggestion
            $item = $nodeList->item(0);
            assert($item instanceof DOMElement);
```

Assertions will break in tests, but get dropped in production

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,198 @@ public function getLockOwner(): ?string
...
             $funding        = new stdClass();
-            $funding->url   = $nodeList->item(0)->getAttribute('url');
-            $funding->title = $nodeList->item(0)->nodeValue;
+            $funding->url   = $item->getAttribute('url');
+            $funding->title = $item->nodeValue;
```
gsteel, line 112: FYI, it is sometime easier to read something like:

```php
$funding = (object) [
    'url' => $item->getAttribute('url'),
    'title' => $item->nodeValue,
];
```

Not asking you to change all of these though 👍

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,198 @@ public function getLockOwner(): ?string
...
+            $object = $this->data['license'];
+            return $object;
+        }
+
+        $license = null;
+
+        $nodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:license');
+
+        if ($nodeList->length > 0) {
+            /** @var DOMElement $item */
```
gsteel: Please swap for an assertion

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,198 @@ public function getLockOwner(): ?string
...
+            $object = $this->data['location'];
+            return $object;
+        }
+
+        $location = null;
+
+        $nodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:location');
+
+        if ($nodeList->length > 0) {
+            /** @var DOMElement $item */
```
gsteel: `assert` instead of `@var` please

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,198 @@ public function getLockOwner(): ?string
...
+            $object = $this->data['images'];
+            return $object;
+        }
+
+        $images = null;
+
+        $nodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:images');
+
+        if ($nodeList->length > 0) {
+            /** @var DOMElement $item */
```
gsteel: `assert` instead of `@var` please

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,198 @@ public function getLockOwner(): ?string
...
+            $object = $this->data['updateFrequency'];
+            return $object;
+        }
+
+        $updateFrequency = null;
+
+        $nodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:updateFrequency');
+
+        if ($nodeList->length > 0) {
+            /** @var DOMElement $item */
```
gsteel: `assert` instead of `@var` please

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,198 @@ public function getLockOwner(): ?string
...
+
+        $this->data['updateFrequency'] = $updateFrequency;
+
+        return $this->data['updateFrequency'];
+    }
+
+    /**
+     * Get the podcast people
+     *
+     * @return list<stdClass|null>
```
gsteel:
```suggestion
     * @return list<object|null>
```

If you add a class-level doc block of:

```php
/**
 * @psalm-type PersonType = object{
 *     name: string,
 *     role: string,
 *     other-prop: string|null|etc…,
 * }
 */
```

Then, you can annotate this method with `@return list<PersonType>|null` and use `PersonType` anywhere relevant: https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-type

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,198 @@ public function getLockOwner(): ?string
...
+     */
+    public function getPodcastIndexPeople(): array
+    {
+        if (array_key_exists('people', $this->data)) {
+            /** @var list<stdClass|null> $people */
```
gsteel: With psalm types, this would need to be `@var list<PersonType>|null $people`

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -56,30 +65,198 @@ public function getLockOwner(): ?string
...
+            $people = $this->data['updateFrequency'];
+            return $people;
+        }
+
+        $nodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:person');
+
+        $personCollection = [];
+
+        if ($nodeList->length) {
+            /** @var DOMElement $entry */
```
gsteel: assertion

src/Writer/Extension/PodcastIndex/Feed.php
froschdesign: The term `person` is defined in the specification and it must also be used here: https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/tags/person.md

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +110,185 @@ public function setPodcastIndexFunding(array $value): Feed
...
+        /** @var array<array<string, mixed>> $this->data['people'] */
+        $this->data['people'][] = $value;
+        return $this;
+    }
+
+    /**
+     * Set a new array of people.
+     * If no argument is passed, it will just remove all existing people.
+     *
+     * @param array<array> $values
```
gsteel: Ideally, the parameter would be more precise here:

At class-level `/** @psalm-import-type PersonType from SomeOtherClass */`

and on this method: `@param list<PersonType>`, or, `array<array-key, PersonType>`

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
```diff
@@ -81,4 +87,124 @@ protected function setFunding(DOMDocument $dom, DOMElement $root): void
...
+        if ($license === null) {
+            return;
+        }
+        $el   = $dom->createElement('podcast:license');
+        $text = $dom->createTextNode($license['identifier']);
+        $el->appendChild($text);
+        $el->setAttribute('url', $license['url']);
+        $root->appendChild($el);
+        $this->called = true;
+    }
```
gsteel: Please can you make all of these new methods `private` - I don't see any reason why we should make these available to inheritors - in fact, it should be unnecessary to inherit from this class at all AFAICT

gsteel (changes requested):
Thanks again @nb823 - just a few more adjustments if possible. Sorry to be picky, but it's _really_ valuable to have precise types and minimal exposure to potential BC issues for future maintenance

src/Writer/Extension/PodcastIndex/Feed.php
nb823: In it's handling as single `person` entry, it is still used here as such. Nevertheless, for the handling of a set of multiple `person` entries, I followed your suggestion to use the more intuitive term `people`, since I did not see the PodcastIndex docs speaking against such an implementation concerning the plural of the term.

src/Reader/Extension/PodcastIndex/Feed.php
nb823: Oh, I was not aware that using assertions is a practiced way to ensure type safety. If you prefer that, I will change it for all the get methods.

src/Reader/Extension/PodcastIndex/Feed.php
nb823: I can indeed change that for all the methods. I just wanted the coding style to stick to the already existing methods at first, just to keep the file (and how to read it) consistent.

src/Writer/Extension/PodcastIndex/Feed.php
nb823: In this case (`setPodcastIndexPeople()`), the users are not supposed to insert an array of objects, but an array of arrays.

This is also an implementation that i continued to keep consistency with the previous PodcastIndex extension methods: Expect arrays as parameters in the set methods, but return objects in the get methods. If both should be standardized to the same data type, I would rather use arrays, since other namespaces of the package already return arrays as well.

src/Reader/Extension/PodcastIndex/Feed.php
nb823: While using array cast, psalm considers the node values to be nullable, which needs additional casting or asserting. But if you prefer that, I will gladly rewrite all these blocks.

This is one of the things I continued mainly for consistency within the class.

src/Writer/Extension/PodcastIndex/Feed.php
gsteel: Requiring object parameters to `setWhatever()` would probably be a pain for users - You can of course make Psalm types for `PersonObject` and `PersonArray`, but I'm happy for you to do whatever makes most sense for consistency here - I'm likely less familiar with this lib than you - regardless, annotating the shapes makes a big difference to type safety for users, general readability and our maintenance burdens, so if these can be included, I'd be grateful :)

nb823:
Thanks again @gsteel for your reviews.
Please feel free to request any further changes that you find necessary or useful. I will gladly do updates to match your best practices.

src/Writer/Extension/PodcastIndex/Feed.php
froschdesign: `people` is for me not _intuitive_ because in the specification it is called person and therefore I expected `persons` as a user – even if I am aware of the plural form `people`. Which means `persons` _is closer_ to the specification for me.

By the way the `persons` is also used in the Podcastindex-org: https://github.com/search?q=org%3APodcastindex-org%20persons&type=code

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -224,12 +235,12 @@ public function getPodcastIndexUpdateFrequency(): object|null
...
      *
-     * @return list<stdClass|null>
+     * @psalm-return list<PersonObject>
      */
     public function getPodcastIndexPeople(): array
     {
         if (array_key_exists('people', $this->data)) {
-            /** @var list<stdClass|null> $people */
+            /** @var list<PersonObject> $people */
             $people = $this->data['updateFrequency'];
```
gsteel: Looks like this array key is incorrect - can we have a test to verify retrieval of People?

```suggestion
            $people = $this->data['people'];
```

nb823:
@froschdesign thank you for linking the feature request https://github.com/laminas/laminas-feed/issues/95 here. Just so you know, this pull request will not resolve the feature request, since this is only the first set of all the requested namespaces to be implemented.

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +122,199 @@ public function setPodcastIndexFunding(array $value): Feed
...
+            );
+        }
+        $this->data['images'] = $value;
+        return $this;
+    }
+
+    /**
+     * Set feed update frequency
+     *
+     * @param array{description: string, complete?: bool, dtstart?: DateTime, rrule?: string} $value
```
gsteel:
```suggestion
     * @param array{description: string, complete?: bool, dtstart?: DateTimeInterface, rrule?: string} $value
```

src/Writer/Extension/PodcastIndex/Feed.php
```diff
@@ -106,6 +122,199 @@ public function setPodcastIndexFunding(array $value): Feed
...
+                'invalid parameter: key "description" of "updateFrequency" must be of type string'
+            );
+        }
+        if (isset($value['complete']) && ! is_bool($value['complete'])) {
+            throw new Writer\Exception\InvalidArgumentException(
+                'invalid parameter: key "complete" of "updateFrequency": must be of type boolean'
+            );
+        }
+        if (isset($value['dtstart'])) {
+            if (! $value['dtstart'] instanceof DateTime) {
```
gsteel: Please use `DateTimeInterface` here

froschdesign:
@nb823 
> …this pull request will not resolve the feature request, since this is only the first set of all the requested namespaces to be implemented.

Correct but we can reopen the related issue.

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
```diff
@@ -49,7 +55,7 @@ protected function _appendNamespaces(): void
     /**
      * Set feed lock
      */
-    protected function setLocked(DOMDocument $dom, DOMElement $root): void
+    private function setLocked(DOMDocument $dom, DOMElement $root): void
```
gsteel: Sorry - this is a BC break for anyone that previously extended this class, however unlikely!

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
```diff
@@ -67,7 +73,7 @@ protected function setLocked(DOMDocument $dom, DOMElement $root): void
     /**
      * Set feed funding
      */
-    protected function setFunding(DOMDocument $dom, DOMElement $root): void
+    private function setFunding(DOMDocument $dom, DOMElement $root): void
```
gsteel: Same BC break here

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
```diff
@@ -81,4 +87,124 @@ protected function setFunding(DOMDocument $dom, DOMElement $root): void
...
+    {
+        /** @psalm-var null|array<string, mixed> $updateFrequency */
+        $updateFrequency = $this->getDataContainer()->getPodcastIndexUpdateFrequency();
+        if ($updateFrequency === null) {
+            return;
+        }
+        $el   = $dom->createElement('podcast:updateFrequency');
+        $text = $dom->createTextNode((string) $updateFrequency['description']);
+        $el->appendChild($text);
+        if (! empty($updateFrequency['complete'])) {
```
gsteel: Didn't catch this before - `empty` is effectively disallowed for new code, please be explicit with comparisons.

I assume in this case, we'd be expecting boolean|unset, so:

```php
if (($updateFrequency['complete'] ?? null) === true) {
    $el->setAttribute('complete', '');
}
```

Can you please remove all further usage of `empty` in this class? Sorry to be a pain.

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
```diff
@@ -81,4 +87,124 @@ protected function setFunding(DOMDocument $dom, DOMElement $root): void
...
         $this->called = true;
     }
+
+    /**
+     * Set feed license
+     */
+    private function setLicense(DOMDocument $dom, DOMElement $root): void
+    {
+        /** @psalm-var null|array<string,string> $license */
+        $license = $this->getDataContainer()->getPodcastIndexLicense();
```
gsteel: What does `$this->getDataContainer()` return?

If this can be asserted, can we get decent types when calling `getPodcastIndexLicense()`?

Same question for all the other "setters" here - will hopefully make the many conditionals less verbose.

gsteel (changes requested):

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
nb823: This will bring back some code that you previously asked to remove, at least concerning the $person and $people arrays. Changing `if(empty($people)` to `if($people === [])`, (which strictly follows the current definition of $people) fired some very new errors during testing that force one to also check on null and hence define those variables as nullable again. So, I hope those changes/reversions will not violate current best practice rules.

psalm-baseline.xml
```diff
@@ -3131,7 +3146,12 @@
   <file src="src/Writer/Extension/PodcastIndex/Renderer/Feed.php">
     <MixedMethodCall>
       <code>getPodcastIndexFunding</code>
+      <code>getPodcastIndexImages</code>
+      <code>getPodcastIndexLicense</code>
+      <code>getPodcastIndexLocation</code>
       <code>getPodcastIndexLocked</code>
+      <code>getPodcastIndexPeople</code>
+      <code>getPodcastIndexUpdateFrequency</code>
```
gsteel: This is the result of Psalm having no idea what `$this->getDataContainer()` returns. I assume it would be a `Reader\Extension\PodcastIndex\Feed` instance??

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
gsteel: Sorry, if I'm going around in circles here - Psalm should be able to infer whatever comes out of `$this->getDataContainer()->getWhatever()` with a little help.

If you focus on ensuring the type you get from `$this->getDataContainer()` is something well-defined, then most of those `empty` calls should be replaceable with `isset` unless I'm just plain wrong?

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
nb823: I made a clearer definition like this:
```
/** @psalm-var FeedWriter $container */
$container = $this->getDataContainer();
```
Since this is redundant for each method, I could alternatively set $container as a class property using a constructor. But since the parent of this class is abstract and without constructor itself that i could call, i would need to fill a lot of properties into the new one. Which way would you prefer?

gsteel: As this is an inherited method, adding `/** @psalm-var Whatever $container */` is fine by me.

Ideally this would be solved with generics on the abstract/parent but that's not a job for this patch :)

I expect the container is lazy loaded, hence the `getDataContainer` method, the only way you could cut down on the repeated `/** @var */` annotations would be to override this method, so I'd just stick to the repeated annotations for now 👍

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
nb823: In fact, this came from another issue, but that one is also solved now.

gsteel (approved):
Thanks again for all your work here @nb823 👍

nb823:
Thank you for your feedback and your patience @gsteel 
The next set of namespaces is already in progress, I hope the next round will take less time from you.

src/Writer/Extension/PodcastIndex/Feed.php
nb823: @froschdesign  Ok, this is in fact crucial. May I add additional alias methods using the term `persons`?

PR #104, opened 2025-09-11
PodcastIndex item and feed

nb823:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

1. Extended PodcastIndex support for additional namespaces:

Feed:
- Feed Remote Item
- Feed Podroll
- Feed Publisher
- Feed Value
- Feed Value Recipient
- Feed Social Interact
- Feed Image

Episode:
- Entry Location
- Entry License
- Entry Person
- Entry Txt
- Entry Social Interact
- Entry Value
- Entry Value Recipient
- Entry Value Time Split
- Entry Season
- Entry Episode
- Entry Alternate Enclosure
- Entry Source
- Entry Integrity
- Entry Image

2. Added help classes within PodcastIndex extension: 'Validator', 'Element Generator' and 'AttributesReader', to be usable by both, feed and entry, especially for shared namespaces
3. Marked namespace 'Images' as deprecated
4. Added tests for the above namespaces
5. Docs: Added documentation for the new namespaces
6. Docs: Added examples for the more complex namespaces 'Value' and 'Alternate Enclosure'
7. Docs: Added alphabetical order by namespace and more info on each set method


froschdesign:
@gsteel 
Here, I would like to ask you to provide an assessment of the classes using only static methods and the general direction.
Thanks in advance! 👍🏻

src/Reader/Extension/PodcastIndex/AttributesReader.php
```diff
@@ -0,0 +1,460 @@
...
+ *        title?: string,
+ *        rel?: string,
+ *        codecs?: string,
+ *        default?: bool,
+ *       sources: list<SourceObject>,
+ *       integrity?: IntegrityObject,
+ *      }
+ */
+class AttributesReader
+{
```
gsteel: As a static helper tool/utility, this class should not be exposed to users, we want to be able to break BC on classes like this whenever we feel like it.

This likely applies to the introduced classes `ElementGenerator` and `Validator` too.

```suggestion
 * }
 *
 * This class is internal to the library and should not be referenced by consumer code.
 * Backwards Incompatible changes can occur in Minor and Patch Releases.
 * 
 * @psalm-internal Laminas\Feed
 * @psalm-internal LaminasTest\Feed
 */
final class AttributesReader
{
```

gsteel (changes requested):
Thanks @nb823 - Helper classes should be aggressively marked as internal so that users are adequately warned not to import them and we don't take on more BC burdens

src/Writer/Extension/PodcastIndex/Validator.php
```diff
@@ -0,0 +1,914 @@
...
+ */
+final class Validator
+{
+    /**
+     * Validate person
+     *
+     * @psalm-param PersonArray $value
+     * @throws Writer\Exception\InvalidArgumentException
+     */
+    public static function validatePerson(array $value): void
```
gsteel, line 171: General question about all the validation methods here: They act like assertions, so I would probably have made the signatures something like:

```php
/**
 * @param array<array-key, mixed> $data
 * @return SomeExplicitArrayShape
 */
public static function validateSomething(array $data): array;
```

So that in your consuming code, you can `$data = Validator::validateSomething($input);` without having to assert or annotate the returned type.

The other way you can approach this is by [annotating the method with `@psalm-assert`](https://psalm.dev/docs/annotating_code/adding_assertions/), i.e

```php
/**
 * @param array<array-key, mixed> $data
 * @psalm-assert SomeExplicitArrayShape $data
 */
public static function validateSomething(array $data): void;
```

At the moment, it looks like you are requiring consuming code to provide valid array shapes prior to validation, that's why you're getting a lot of `DocblockTypeContradiction` issues with Psalm. Psalm is telling you that why are you checking `isset($value['name'])` when `ExplicitArrayShape` is `array{name: string}`

src/Reader/Extension/PodcastIndex/AttributesReader.php
```diff
@@ -0,0 +1,464 @@
...
+        $license->identifier = $item->nodeValue;
+        $license->url        = $item->getAttribute('url');
+
+        return $license;
+    }
+
+    /**
+     * Read podcast location
+     *
+     * @psalm-param DOMElement $item
```
gsteel: On all of these methods `@psalm-param` is unnecessary.

Also there's no need to vendor prefix the `@param` - PHPStan, PHPStorm and Psalm will use `@param SomeShape $value` just fine

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -5,35 +5,56 @@
...
+ * @psalm-import-type SourceObject from AttributesReader
+ * @psalm-import-type IntegrityObject from AttributesReader
+ * @psalm-import-type AlternateEnclosureObject from AttributesReader
  */
 class Entry extends Extension\AbstractEntry
 {
     /**
      * Get the entry transcript
-     *
-     * @psalm-return null|object{url: string, type: string, language: string, rel: string}
```
gsteel, line 22: Why not `@return TranscriptObject` here?

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -5,35 +5,56 @@
...
      */
     public function getTranscript(): ?stdClass
     {
         if (array_key_exists('transcript', $this->data)) {
+            /** @psalm-var stdClass */
```
gsteel: I understand why you are _forcing_ the type here, but you could instead override the `$this->data` type annotation with something like

```
/**
 * @psalm-type array{
 *     transcript?: TranscriptObject,
 *     // Other specific array keys
 * }
 */
protected $data; // Whatever fits the parent class definition
```

You may get an issue from Psalm such as `NonInvariantPropertyType` which can be supressed, but you'll end up with a lot less assertions and annotations as a result

gsteel (changes requested):
Thanks @nb823 - I haven't the time for a full review but I've added a few things that I spotted 👍

src/Reader/Extension/PodcastIndex/Entry.php
nb823: Because of a mismatch with the return type `?StdClass`:

`psalm: MismatchingDocblockReturnType: Docblock has incorrect return type 'null|type-alias(Laminas\Feed\Reader\Extension\PodcastIndex\AttributesReader::TranscriptObject)', should be 'null|stdClass'`

I will need to change the return type to `object|null` if I want to keep the Docblock as above.

nb823: This was old code, not written by myself, but throwing new psalm errors that i had to solve. So my priority was to keep the defined return type as it is.

src/Reader/Extension/PodcastIndex/Entry.php
nb823: So far, I did not find a way to implement your suggestion without causing a lot of new psalm errors and additional problems to solve. Also when assigning `@psalm-var `instead of `@psalm-type`. I would need to add a constructor, and still it would throw new `InvalidPropertyAssignmentValue` errors wherever I am using variables to fill the `$this->data` entries.

src/Reader/Extension/PodcastIndex/Entry.php
gsteel: In this case, the method _is_ actually returning `TranscriptObject|null`, so the return type annotation should be used to override the native type and the Psalm issue suppressed until it is safe to change the native return type, i.e. in the next major.

The improved type inference wins over fewer baselined Psalm issues if that makes sense.

src/Reader/Extension/PodcastIndex/Entry.php
gsteel: OK, I can imagine the avalanche of Psalm issues that would occur here.

I made a mistake in the code example, it should be `@var`, not `@psalm-type`.

Possibly, in the class hierarchy, finding the `parent::$data` declaration and changing that to `@var array<string, mixed>`, then narrowing in child classes may work, but I understand reluctance to do this due to the sheer volume of issues it could produce.

nb823: Trying this approach, I could not reduce the issues mentioned above, so I would rather stick to the specific annotations in each method.

nb823:
While this PR was open, I have also been working on the last bundle of updates on the PodcastIndex extension. That one is almost finished. Should I close this PR and open a new one whenever the final updates are ready?

gsteel:
> While this PR was open, I have also been working on the last bundle of updates on the PodcastIndex extension. That one is almost finished. Should I close this PR and open a new one whenever the final updates are ready?

@nb823 Please don't close here. This is a big patch and there is a lot to review, so please don't make it bigger by combining it with more work…

I'll try to fit in a review ASAP.

Can I assume that we should wait to release the next minor if you have more patches incoming?

nb823:
> Can I assume that we should wait to release the next minor if you have more patches incoming?

@gsteel Yes, I think it makes sense to wait until then.
I will also wait with opening the last PR until this one's review is finished.

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -46,24 +70,38 @@ public function getTranscript(): ?stdClass
         return $this->data['transcript'];
     }
 
+    /**
+     * Get the entry transcript
+     *
+     * @psalm-suppress MismatchingDocblockReturnType
+     * @return null|TranscriptObject
+     */
+    public function getPodcastIndexTranscript(): ?stdClass
```
gsteel: New code should not return `stdClass` and prefer `object`. This allows for future narrowing of the type internally without breaking BC. i.e if we made a concrete `Transcript` object, we could return that without the need to change the return type. Hinting on `stdClass` means the returned type must inherit from `stdClass`

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -46,24 +70,38 @@ public function getTranscript(): ?stdClass
...
+    {
+        /** @psalm-var stdClass */
```
gsteel: forcing this var should be unnecessary

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -74,15 +112,27 @@ public function getChapters(): ?stdClass
         return $this->data['chapters'];
     }
 
+    /**
+     * Get the entry chapters
+     *
+     * @psalm-suppress MismatchingDocblockReturnType
+     * @return null|ChaptersObject
+     */
+    public function getPodcastIndexChapters(): ?stdClass
```
gsteel: prefer `object` over `stdClass`

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -5,35 +5,59 @@
...
      * Get the entry transcript
      *
-     * @psalm-return null|object{url: string, type: string, language: string, rel: string}
+     * @psalm-suppress MismatchingDocblockReturnType
+     * @return null|TranscriptObject
      */
     public function getTranscript(): ?stdClass
     {
         if (array_key_exists('transcript', $this->data)) {
+            /** @psalm-var stdClass */
```
gsteel: If anything, this should be `/** @psalm-var TranscriptObject */`

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -46,24 +70,38 @@ public function getTranscript(): ?stdClass
         return $this->data['transcript'];
     }
 
+    /**
+     * Get the entry transcript
+     *
+     * @psalm-suppress MismatchingDocblockReturnType
```
gsteel: Please remove this suppression. Generally speaking inline supressions should not be used in `/src` and instead added to the baseline file.

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -46,24 +70,38 @@ public function getTranscript(): ?stdClass
...
+    {
+        /** @psalm-var stdClass */
+        return $this->getTranscript();
+    }
+
     /**
      * Get the entry chapters
      *
-     * @psalm-return null|object{url: string, type: string}
+     * @psalm-suppress MismatchingDocblockReturnType
```
gsteel: Please baseline

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -46,24 +70,38 @@ public function getTranscript(): ?stdClass
...
+     * @return null|ChaptersObject
      */
     public function getChapters(): ?stdClass
     {
         if (array_key_exists('chapters', $this->data)) {
+            /** @psalm-var stdClass */
```
gsteel: Should be `ChaptersObject`

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -74,15 +112,27 @@ public function getChapters(): ?stdClass
         return $this->data['chapters'];
     }
 
+    /**
+     * Get the entry chapters
+     *
+     * @psalm-suppress MismatchingDocblockReturnType
```
gsteel: Please remove the supression

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -74,15 +112,27 @@ public function getChapters(): ?stdClass
...
+     * @return null|ChaptersObject
+     */
+    public function getPodcastIndexChapters(): ?stdClass
+    {
+        /** @psalm-var stdClass */
```
gsteel: unnecessary

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -107,6 +157,387 @@ public function getSoundbites(): array
...
+        }
+
+        $values         = [];
+        $valuesNodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:value');
+
+        foreach ($valuesNodeList as $valueNode) {
+            assert($valueNode instanceof DOMElement);
+            $valueObject = AttributesReader::readValue($valueNode);
+
+            /** @psalm-suppress TooManyArguments */
```
gsteel: This supression is odd. `DOMXPath::query` takes up to 3 args.

Please remove (It should be in the baseline if necessary), and also check the documented type for `$this->xpath`

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -107,6 +157,387 @@ public function getSoundbites(): array
...
+            $valueRecipients         = [];
+
+            foreach ($valueRecipientsNodeList as $entry) {
+                assert($entry instanceof DOMElement);
+                $object            = AttributesReader::readValueRecipient($entry);
+                $valueRecipients[] = $object;
+            }
+            $valueObject->valueRecipients = $valueRecipients;
+
+            /** @psalm-suppress TooManyArguments */
```
gsteel: Please remove

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -107,6 +157,387 @@ public function getSoundbites(): array
...
+    /**
+     * Get value time split
+     *
+     * @return ValueTimeSplitObject
+     */
+    private function getValueTimeSplit(DOMElement $entry): object
+    {
+        $object = AttributesReader::readValueTimeSplit($entry);
+
+        /** @psalm-suppress TooManyArguments */
```
gsteel: Please remove

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -107,6 +157,387 @@ public function getSoundbites(): array
...
+        $itemsNodeList = $this->xpath->query('podcast:remoteItem', $entry);
+        if ($itemsNodeList->length > 0) {
+            assert($itemsNodeList[0] instanceof DOMElement);
+            $itemsObject        = AttributesReader::readRemoteItem($itemsNodeList[0]);
+            $object->remoteItem = $itemsObject;
+        }
+
+        /** @psalm-suppress TooManyArguments */
```
gsteel: Please remove

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -107,6 +157,387 @@ public function getSoundbites(): array
...
+        }
+
+        $enclosures         = [];
+        $enclosuresNodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:alternateEnclosure');
+
+        foreach ($enclosuresNodeList as $enclosureNode) {
+            assert($enclosureNode instanceof DOMElement);
+            $enclosureObject = AttributesReader::readAlternateEnclosure($enclosureNode);
+
+            /** @psalm-suppress TooManyArguments */
```
gsteel: Please remove

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -107,6 +157,387 @@ public function getSoundbites(): array
...
+            if ($sourcesNodeList->length > 0) {
+                foreach ($sourcesNodeList as $entry) {
+                    assert($entry instanceof DOMElement);
+                    $object    = AttributesReader::readSource($entry);
+                    $sources[] = $object;
+                }
+                $enclosureObject->sources = $sources;
+            }
+
+            /** @psalm-suppress TooManyArguments */
```
gsteel: Please remove

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -107,6 +157,387 @@ public function getSoundbites(): array
...
+                assert($entry instanceof DOMElement);
+                $image    = AttributesReader::readDetailedImage($entry);
+                $images[] = $image;
+            }
+        }
+
+        $this->data['detailedImages'] = $images;
+
+        return $this->data['detailedImages'];
+    }
```
gsteel: IMO, it would be better to remove `null` from the return type and return `[]` where neccessary

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -210,15 +194,44 @@ public function getPodcastIndexImages(): object|null
...
         return $this->data['images'];
     }
 
+    /**
+     * Get the podcast detailed images.
+     * Returns the contents of one or more `<podcast:image>` tags.
+     *
+     * @psalm-return null|list<DetailedImageObject>
+     */
+    public function getPodcastIndexDetailedImages(): null|array
```
gsteel: Can we remove `null` from the return type here?

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -253,12 +262,12 @@ public function getPodcastIndexUpdateFrequency(): object|null
     /**
      * Get the podcast people
      *
-     * @psalm-return list<PersonObject>
+     * @psalm-return null|list<PersonObject>
      */
-    public function getPodcastIndexPeople(): array
+    public function getPodcastIndexPeople(): null|array
```
gsteel: Same here with `null` in the return type union. Can it be an empty array?

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -285,6 +289,16 @@ public function getPodcastIndexPeople(): array
...
 
+    /**
+     * Get the podcast persons (alias of getPodcastIndexPeople)
+     *
+     * @psalm-return null|list<PersonObject>
+     */
+    public function getPodcastIndexPersons(): null|array
+    {
+        return $this->getPodcastIndexPeople();
+    }
```
gsteel: `null` here too?

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -377,12 +383,12 @@ public function getPodcastIndexMedium(): object|null
     /**
      * Get the podcast blocks
      *
-     * @return list<object{value: string, id?: string}>
+     * @return null|list<object{value: string, id?: string}>
      */
-    public function getPodcastIndexBlocks(): array
+    public function getPodcastIndexBlocks(): null|array
```
gsteel: If this has been released, adding `null` here should be a BC break, again, what's wrong with `[]` instead?

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -407,12 +410,12 @@ public function getPodcastIndexBlocks(): array
     /**
      * Get the podcast txts
      *
-     * @return list<object{value: string, purpose?: string}>
+     * @return null|list<TxtObject>
      */
-    public function getPodcastIndexTxts(): array
+    public function getPodcastIndexTxts(): null|array
```
gsteel: Same comment about `null`

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -464,6 +462,168 @@ public function getPodcastIndexPodping(): object|null
         return $this->data['podping'];
     }
 
+    /**
+     * Get the podcast remoteItems
+     *
+     * @return null|list<RemoteItemObject>
+     */
+    public function getPodcastIndexRemoteItems(): null|array
```
gsteel: Again `null`

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -464,6 +462,168 @@ public function getPodcastIndexPodping(): object|null
...
+        }
+
+        $podrollItems    = [];
+        $podrollNodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:podroll');
+
+        if ($podrollNodeList->length > 0) {
+            $podrollNode = $podrollNodeList->item(0);
+            assert($podrollNode instanceof DOMElement);
+
+            /** @psalm-suppress TooManyArguments */
```
gsteel: Please remove

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -464,6 +462,168 @@ public function getPodcastIndexPodping(): object|null
...
+            return $publisherItem;
+        }
+
+        $publisherItem     = null;
+        $publisherNodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:publisher');
+
+        if ($publisherNodeList->length > 0) {
+            $publisherNode = $publisherNodeList->item(0);
+            assert($publisherNode instanceof DOMElement);
+            /** @psalm-suppress TooManyArguments */
```
gsteel: Please remove

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -464,6 +462,168 @@ public function getPodcastIndexPodping(): object|null
...
+
+        return $this->data['remoteItems'];
+    }
+
+    /**
+     * Get the podcast podroll remote items
+     *
+     * @return null|list<RemoteItemObject>
+     */
+    public function getPodcastIndexPodroll(): null|array
```
gsteel: `null|array` -> `array`

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -464,6 +462,168 @@ public function getPodcastIndexPodping(): object|null
...
+
+        return $this->data['publisher'];
+    }
+
+    /**
+     * Get the podcast values
+     *
+     * @return null|list<ValueObject>
+     */
+    public function getPodcastIndexValues(): null|array
```
gsteel: `null|array` -> `array`

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -464,6 +462,168 @@ public function getPodcastIndexPodping(): object|null
...
+        }
+
+        $values         = [];
+        $valuesNodeList = $this->xpath->query($this->getXpathPrefix() . '/podcast:value');
+
+        foreach ($valuesNodeList as $valueNode) {
+            assert($valueNode instanceof DOMElement);
+            $valueObject = AttributesReader::readValue($valueNode);
+
+            /** @psalm-suppress TooManyArguments */
```
gsteel: Please remove

src/Reader/Extension/PodcastIndex/Feed.php
```diff
@@ -464,6 +462,168 @@ public function getPodcastIndexPodping(): object|null
...
+
+        return $this->data['values'];
+    }
+
+    /**
+     * Get the podcast social interacts
+     *
+     * @return null|list<SocialInteractObject>
+     */
+    public function getPodcastIndexSocialInteracts(): null|array
```
gsteel: `null|array` -> `array`

src/Writer/Extension/PodcastIndex/Validator.php
```diff
@@ -0,0 +1,942 @@
...
+                'invalid parameter: key "purpose" of "txt" must be of type string'
+            );
+        }
+    }
+
+    /**
+     * Validates images srcset
+     *
+     * @param array<array-key, mixed> $value
+     * @psalm-assert array{srcset: string} $value
```
gsteel: `ImagesArray`?

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -54,20 +81,15 @@ protected function _appendNamespaces(): void
      */
     protected function setTranscript(DOMDocument $dom, DOMElement $root): void
     {
-        /** @psalm-var null|array<string, string> $locked */
-        $locked = $this->getDataContainer()->getPodcastIndexTranscript();
-        if ($locked === null) {
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
+
+        /** @psalm-var null|TranscriptArray $transcript */
```
gsteel, line 96: Forcing the var here should be superfluous - it should already be in the return type of `getPodcastIndexTranscript`

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -77,14 +99,15 @@ protected function setTranscript(DOMDocument $dom, DOMElement $root): void
      */
     protected function setChapters(DOMDocument $dom, DOMElement $root): void
     {
-        /** @psalm-var null|array<string, string> $chapters */
-        $chapters = $this->getDataContainer()->getPodcastIndexChapters();
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
+
+        /** @psalm-var null|ChaptersArray $chapters */
```
gsteel, line 113: Unnecessary type forcing?

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -94,22 +117,261 @@ protected function setChapters(DOMDocument $dom, DOMElement $root): void
      */
     protected function setSoundbites(DOMDocument $dom, DOMElement $root): void
     {
-        /** @psalm-var null|list $soundbites */
-        $soundbites = $this->getDataContainer()->getPodcastIndexSoundbites();
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
+
+        /** @psalm-var null|list<SoundbiteArray> $soundbites */
```
gsteel, line 130: Unnecessary type forcing?

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -94,22 +117,261 @@ protected function setChapters(DOMDocument $dom, DOMElement $root): void
...
+
+    /**
+     * Set entry location
+     */
+    private function setLocation(DOMDocument $dom, DOMElement $root): void
+    {
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
+
+        /** @psalm-var null|LocationArray $location */
```
gsteel, line 149: Unnecessary type forcing?

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -94,22 +117,261 @@ protected function setChapters(DOMDocument $dom, DOMElement $root): void
...
         }
     }
+
+    /**
+     * Set entry location
+     */
+    private function setLocation(DOMDocument $dom, DOMElement $root): void
+    {
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
```
gsteel: You force this type every time. It's OK to create a private method:

```php
private function getEntryWriter(): EntryWriter
{
    $container = $this->getDataContainer();
    assert($container instanceof EntryWriter);

    return $container;
}
```

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -94,22 +117,261 @@ protected function setChapters(DOMDocument $dom, DOMElement $root): void
...
+
+    /**
+     * Set feed license
+     */
+    private function setLicense(DOMDocument $dom, DOMElement $root): void
+    {
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
+
+        /** @psalm-var null|LicenseArray $license */
```
gsteel, line 166: Unnecessary type forcing?

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -94,22 +117,261 @@ protected function setChapters(DOMDocument $dom, DOMElement $root): void
...
+
+    /**
+     * Set feed people
+     */
+    private function setPeople(DOMDocument $dom, DOMElement $root): void
+    {
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
+
+        /** @psalm-var null|list<PersonArray> $people */
```
gsteel, line 184: Unnecessary type forcing?

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -94,22 +117,261 @@ protected function setChapters(DOMDocument $dom, DOMElement $root): void
...
+
+    /**
+     * Set entry txts
+     */
+    private function setTxts(DOMDocument $dom, DOMElement $root): void
+    {
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
+
+        /** @psalm-var list<TxtArray>|null $txts */
```
gsteel, line 203: Unnecessary type forcing?

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -94,22 +117,261 @@ protected function setChapters(DOMDocument $dom, DOMElement $root): void
...
+        $txts = $container->getPodcastIndexTxts();
+        if ($txts === null || $txts === []) {
+            return;
+        }
```
gsteel, line 207: Case in point, if `getPodcastIndexTxts` only returns an array, you don't need to check for null here

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -94,22 +117,261 @@ protected function setChapters(DOMDocument $dom, DOMElement $root): void
...
+
+    /**
+     * Set feed social interacts
+     */
+    private function setSocialInteracts(DOMDocument $dom, DOMElement $root): void
+    {
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
+
+        /** @psalm-var list<SocialInteractArray>|null $socialInteracts */
```
gsteel, line 223: Unnecessary type forcing?

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
```diff
@@ -94,22 +117,261 @@ protected function setChapters(DOMDocument $dom, DOMElement $root): void
...
+
+    /**
+     * Set the values
+     */
+    private function setValues(DOMDocument $dom, DOMElement $root): void
+    {
+        /** @psalm-var EntryWriter $container */
+        $container = $this->getDataContainer();
+
+        /** @psalm-var list<ValueArray>|null $values */
```
gsteel, line 244: Unnecessary type forcing?

src/Writer/Extension/PodcastIndex/Entry.php
```diff
@@ -114,8 +140,29 @@ public function addPodcastIndexSoundbites(array $values): Entry
...
     /**
-     * Add entry soundbite
+     * Set entry soundbites.
+     * If no argument is passed, the existing soundbite entries get removed.
      *
+     * @param list<SoundbiteArray> $values
+     * @return $this
+     * @throws Writer\Exception\InvalidArgumentException
+     */
+    public function setPodcastIndexSoundbites(array $values = []): Entry
```
gsteel: Returning `self` here makes more sense. I understand you can't change existing methods, but for added code `self` is better.

src/Writer/Extension/PodcastIndex/Entry.php
```diff
@@ -149,6 +196,337 @@ public function addPodcastIndexSoundbite(array $value): Entry
...
+        return $this->setPodcastIndexPeople($values);
+    }
+
+    /**
+     * Add entry txt
+     *
+     * @param TxtArray $value
+     * @return $this
+     * @throws Writer\Exception\InvalidArgumentException
+     * @psalm-suppress DocblockTypeContradiction
```
gsteel: This `DocblockTypeContradiction` should be fixed, not supressed.

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
```diff
@@ -113,15 +123,13 @@ private function setLicense(DOMDocument $dom, DOMElement $root): void
         /** @psalm-var FeedWriter $container */
         $container = $this->getDataContainer();
 
-        /** @psalm-var null|array{identifier: string, url: string} $license */
+        /** @psalm-var null|LicenseArray $license */
```
gsteel, line 133: This type should be known via the return type of `FeedWriter::getPodcastIndexLicense()` right?

gsteel: If the method call is magic - inference would be improved by either adding the concrete method, or adding `@method null|LicenseArray getPodcastIndexLicense()` to the target class

gsteel (changes requested):
Thanks @nb823 - I may be a bit overzealous with the types - and potentially misunderstanding usage of `__call`, but overall, there's a lot of type forcing going on where you've put loads of effort into the type declarations but Psalm is not making the most usage of those types.

src/Reader/Extension/PodcastIndex/Entry.php
nb823: Yes, this is an odd suppression for an odd error:
`Too many arguments for DOMXPath::query - expecting 1 but saw 2`
The docs for `DOMXPath::query` clearly allow up to three arguments.
But I will remove all inline suppressions and move them to the baseline, as you suggested.

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
nb823: Unfortunately not. Here, the writer's get method is called, which is not specifically defined but proxied via the `_call()` method. So, psalm does not find the method and hence does not know it's return type.

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
nb823: The EntryWriter get methods do return null if the respective `$data` key is not set (see the `_call()` method). So, in this case we do need to check on null responses. The empty-array-responses that you suggested further above referred to the EntryReader get methods that are to be called from "outside".

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
nb823: The problem is that type aliases (`@psalm-type` / `@psalm-import-type`) like `LicenseArray` don’t resolve inside `@method` tags. And, in my view, adding the concrete methods (which would be necessary for all get methods of the writer classes) would make the originally introduced magic calls pointless.

src/Writer/Extension/PodcastIndex/Renderer/Entry.php
nb823: Thanks for the suggestion. I also implemented this private method in `src/Writer/Extension/PodcastIndex/Renderer/Feed.php`, since there I had the same issue.

src/Reader/Extension/PodcastIndex/Entry.php
gsteel: What is weirder is that I can't reproduce the issue:

https://psalm.dev/r/30f5e78af9

src/Writer/Extension/PodcastIndex/Renderer/Feed.php
gsteel: The magic is the problem here then. I hate `__call`… I'm not suggesting that you backfill all these getters though, therefore the type-forcing makes sense to me now 👍 

In an ideal world, we'd have something like:

```php

final readonly class Licence {
    public function __construct(
        public string $url,
        public int $whatever,
    ) {}
}

final class FeedOrWriterOrReader {
    public function getLicence(): Licence|null
    {
        return $this->fetch('licence', Licence::class);
    }

    /**
     * @param class-string<T> $type
     * @return T
     * @template T of object
     */
    private function fetch(string $key, string $type): object|null
    {
        $item = $this->data[$key] ?? null;
        if ($item === null) {
            return null;
        }
        assert($item instanceof $type);

        return $item;
    }
}
```

Using something like Valinor, or another internal mechanism to map the expected XML attribs/elements into objects

nb823: Is this a request to refactor the mapping, or can i leave the type-forcing as it is for now?

gsteel: No, please stick with forcing the types for now.

I think that a refactor for the next major might include getting rid of magic, but not in this patch.

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -159,12 +153,9 @@ public function getSoundbites(): array
 
     /**
      * Get the entry soundbites
-     *
-     * @return list<SoundbiteObject>
```
gsteel: The return type should be retained here, otherwise consumers are getting `array<array-key, mixed>`

gsteel (changes requested):
Just one nitpick from me @nb823 

Thanks for all your hard work here.

I could do with another pair of eyes on this patch as it's pretty big :) @froschdesign - Would you oblige?

froschdesign:
> I could do with another pair of eyes on this patch as it's pretty big :) @froschdesign - Would you oblige?

It really is a big pull request, but I like the validator and the generator. Otherwise, the existing classes would have to be extended much more, which would be a nightmare.

froschdesign (approved):
Thanks for all the work! 👍🏻 ❤️ 

(I think we need a blog post to make people aware of the support. We must also update the repository/component description and extend the homepage of the documentation to add this feature.)

gsteel:
Thanks again @nb823 I won't release this yet as I understand you have another PR incoming (?)

nb823:
Thanks for all your reviews and feedback @froschdesign and @gsteel!
Yes, I will create the next (and last) PR as soon as possible, so you can wait with the release.

PR #105, opened 2025-10-08
Podcastindex Live Item

nb823:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | yes
| QA            | yes

### Description

This is again a bigger patch, this time because of the implementation of a new `LiveItem` API within the PodcastIndex extension. I hope this step does not raise too many issues, and I am thankful for your feedback on this solution.

**1. Extended PodcastIndex support for additional namespaces**
* Feed: LiveItem
* Feed: Chat
* Entry: Chat
* Entry: Funding
* Entry: ContentLink

**2. For namespace LiveItem: created new container classes**
* new LiveItem writer, reader and renderer
* all these classes inherit from the respective Entry classes
* goal: LiveItem shall be able to contain all elements that Entry can contain: https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/tags/live-item.md
=> I would be happy about your feedback on the way I implemented this inheritance structure.

**3. Updates on namespace Funding**
* The funding namespace was already implemented for the Feed container, but as a **single** tag. It is passed to the feed writer as a `$data['funding']` value, using the method `setFunding()`.
* This has been updated in PodcastIndex: Now, multiple funding tags are allowed, for the feed and for each entry.
* So, I have added a `$data['fundings']` key that is filled using the new methods `setFundings()` and `addFunding()`.
* The old method is marked as deprecated, but still usable.
* The old getter method `getFunding()` still fetches only the single entry as before, or the first entry if multiple exist in the rss feed.  The new getter `getPodcastIndexFundings()` fetches all existing funding entries, no matter what setter was used before.
* This way, as a user, sticking to the old methods should still work just as before, but users get encouraged to switch to the new methods.
* The funding namespace is newly implemented for the Entry API, but only with the new methods allowing multiple tags.
=> I would be happy about your feedback on this solution.

**4. Added tests for the new methods and for the new LiveItem API**
**5. Added documentation on the new methods and the new LiveItem API**


nb823:
Sorry, I just converted this PR into a draft - let me just add one last change: 
The same update that was done to the `funding` tag by the PodcastIndex project, has also been made to the `location` tag recently. My last PR #104 has just introduced the  `location` namespace for the `Entry` API, but in the now deprecated way. But since it has not been deployed yet, it makes a lot of sense to update that implementation just now, so we have one less deprecated namespace published within the laminas-feed package.

nb823:
I am going to update the `location` namespace of the `Feed` API in the same way I did it to the `funding` namespace, as described in the opening comment of this PR. I hope you approve of the way I did it.

froschdesign:
@nb823 
What is the current status? Is anything still missing?

nb823:
In fact, there is one more thing to do. The introduction of the dynamic ElementGenerator helper made it possible to write unexpected data into the tags. Before, the writer classes saved anything that you would pass to their methods, but only the expected inputs were rendered by the renderer classes. Now, anything that was passed is being rendered, as long as it fits any of the supported data types. The plan: Writer classes (by using the Validator) remove unexpected data (array entries whose keys do not match any of the expected keys) before internally saving and passing it to the renderer classes.

I will set the PR back to "Ready for review" after pushing this update here. And I will leave another comment to make clear whenever this is ready to be merged.

froschdesign:
@nb823 
Can we release [version 2.26.0](https://github.com/laminas/laminas-feed/milestone/36?closed=1) or do you think it would be better to include this pull request?

nb823:
@froschdesign The update will be ready within the next week. But it is not neccessary to wait with the release until then. Sorry for the delay.

froschdesign:
@nb823 
> Sorry for the delay.

No problem at all. Take your time. We are grateful for your help! 👍🏻

nb823:
@froschdesign 
I need to correct myself: Since you did already merge my last PR, it does make sense to wait until this one is finished. I will push the updates ASAP.

nb823:
The update is finished. So now this PR is ready to be reviewed and merged.

src/Reader/Extension/PodcastIndex/LiveItem.php
```diff
@@ -0,0 +1,114 @@
...
+use DOMElement;
+use Laminas\Feed\Reader\Entry\Rss as EntryReader;
+
+use function array_key_exists;
+use function assert;
+
+/**
+ * Describes PodcastIndex LiveItem data in a RSS Feed
+ */
+class LiveItem extends EntryReader
```
froschdesign: Can this class be closed via final?

src/Writer/Extension/PodcastIndex/LiveItem.php
```diff
@@ -0,0 +1,96 @@
...
+namespace Laminas\Feed\Writer\Extension\PodcastIndex;
+
+use Laminas\Feed\Writer\Entry;
+
+/**
+ * Describes LiveItem data in a RSS Feed
+ *
+ * @psalm-import-type LiveItemArray from Validator
+ */
+class LiveItem extends Entry
```
froschdesign: Can this class be closed via final?

src/Reader/Extension/PodcastIndex/AttributesReader.php
```diff
@@ -146,9 +150,19 @@
...
+ *     }
+ * @psalm-type ContentLinkObject = object{
+ *        href: string,
+ *        description: string,
+ *     }
+ * @psalm-type ChatObject = object{
+ *        server: string,
+ *        protocol: string,
+ *        accountId?: string,
+ *        space?: string,
```
froschdesign: Is this the definition with `null` correct? The return type of `DOMElement::getAttribute` says something different:

> The value of the attribute, or an empty string if no attribute with the given qualifiedName is found.

https://www.php.net/manual/domelement.getattribute.php

froschdesign (changes requested):

src/Reader/Extension/PodcastIndex/LiveItem.php
nb823: I think nothing speaks against this. I will set all three LiveItem classes to final.

src/Reader/Extension/PodcastIndex/AttributesReader.php
nb823: You are right, the objects' attributes are never null, unlike the respective arrays' attributes. I fixed this.

froschdesign (approved):

src/Reader/Extension/PodcastIndex/Entry.php
```diff
@@ -160,30 +163,32 @@ public function getPodcastIndexSoundbites(): array
...
 
     /**
-     * Get the entry location
+     * Get the episode locations
      *
-     * @return null|LocationObject
+     * @psalm-return list<LocationObject>
      */
-    public function getPodcastIndexLocation(): object|null
+    public function getPodcastIndexLocations(): array
```
gsteel, line 170: Note to self: Not a BC break because it has not been released

src/Writer/Extension/PodcastIndex/Validator.php
```diff
@@ -6,14 +6,21 @@
 
 use DateTimeInterface;
 use Laminas\Feed\Writer;
+use LockedArray;
+use SeasonArray;
```
gsteel: These imports look suspect - did your IDE do this for locally defined Psalm types?

src/Writer/Extension/PodcastIndex/Validator.php
```diff
@@ -156,19 +171,106 @@
...
+ * @psalm-type ContentLinkArray = array{
+ *         href: string,
+ *         description: string,
+ *       }
+ * @psalm-type ChatArray = array{
+ *         server: string,
+ *         protocol: string,
+ *         accountId?: string,
+ *         space?: string,
+ *       }
```
gsteel, line 191: Not a big deal, but I've since realised that whilst `@psalm-internal Namespace` means that _our_ tests and code don't issue errors on use, we should also add `@internal` for users as not all IDEs support the psalm equivalent.

src/Writer/Extension/PodcastIndex/Entry.php
```diff
@@ -167,46 +156,44 @@ public function setPodcastIndexSoundbites(array $values = []): self
...
 
     /**
-     * Set entry location
+     * Adds a feed location tag.
      *
      * @param LocationArray $value
      * @return $this
-     * @throws Writer\Exception\InvalidArgumentException
      */
-    public function setPodcastIndexLocation(array $value): self
```
gsteel, line 206: Not a BC Break - previous version not released

gsteel (changes requested):
This patch is massive! Thanks for all your hard work.

Apart from the stray imports, this looks OK to me.

Can you have another pass at the added classes and make sure that anything that can be made `@internal` is - this is a huge expansion of the api, so anything we can do to reduce it is a good thing.

nb823:
Thanks for your reviews.

> Can you have another pass at the added classes and make sure that anything that can be made `@internal` is - this is a huge expansion of the api, so anything we can do to reduce it is a good thing.

I think it's only the three helper classes that should be internal. I marked them accordingly.

gsteel (changes requested):
Internal comments should take the form:

```php
/**
 * Description
 *
 * @internal
 *
 * @psalm-internal Some\Ns
 * @psalm-internal Other\Ns
 */
```

i.e. bare `@internal` cannot accept a namespace argument

Coding standards require certain order/grouping of comments, so run `phpcbf` if you get any problems 👍

gsteel (approved):
Thanks again @nb823 👍

gsteel:
@nb823
Do you any more changes to make or can we release `2.26.0` ??

nb823:
@gsteel 
No more changes, you can release it :)
Thank you for all your detailed reviews.

PR #116, opened 2026-03-13
update to psalm ^6.15.1, composer bump --dev-only

fezfez:
fix #111 , #110, #90


gsteel (approved):
Phew! That's a pretty big baseline 😬. Thanks @fezfez

PR #119, opened 2026-03-28
Remove redundant processing of an invalid date field

### LINKED ISSUE #78: DateTime::createFromFormat(): Argument #2 ($datetime) must be of type string, bool given
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.21.0

#### Summary

DateTime::createFromFormat(): Argument #2 ($datetime) must be of type string, bool given {"exception":"[object] (TypeError(code: 0): DateTime::createFromFormat(): Argument #2 ($datetime) must be of type string, bool given at /var/www/html/releases/2/vendor/laminas/laminas-feed/src/Reader/Feed/Rss.php:285)

#### Current behavior

image

#### How to reproduce

I am searching for the feed it failed, will post the feed url later!

#### Expected behavior

Should not throw exception.

bramley:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This change removes redundant further processing  when an invalid date field has already failed. That removes the cause of the original issue #78 which was that an incorrect variable was being used.

The date parser used by strtotime, DateTime etc already handles dates in the RSS, RFC822, RFC2822 formats so there is no purpose in validating against those specific formats.
https://www.php.net/manual/en/datetime.formats.php#datetime.formats.compound

Unit tests before and after the change have the same result

```
TESTS_LAMINAS_FEED_READER_ONLINE_ENABLED=1 phpunit --group Laminas_Feed --verbose

OK, but incomplete, skipped, or risky tests!
Tests: 1536, Assertions: 1760, Skipped: 2, Incomplete: 11.
```

### DIFF
--- a/src/Reader/Entry/Rss.php
+++ b/src/Reader/Entry/Rss.php
@@ -179,7 +179,7 @@ class Rss extends AbstractEntry implements EntryInterface
     /**
      * Get the entry's date of creation
      *
-     * @return DateTime
+     * @return null|DateTime
      */
     public function getDateCreated()
     {
@@ -189,7 +189,7 @@ class Rss extends AbstractEntry implements EntryInterface
     /**
      * Get the entry's date of modification
      *
-     * @return DateTime
+     * @return null|DateTime
      * @throws Exception\RuntimeException
      */
     public function getDateModified()
@@ -206,31 +206,15 @@ class Rss extends AbstractEntry implements EntryInterface
         ) {
             $dateModified = $this->xpath->evaluate('string(' . $this->xpathQueryRss . '/pubDate)');
             if ($dateModified) {
-                $dateModifiedParsed = strtotime($dateModified);
-                if ($dateModifiedParsed) {
-                    $date = new DateTime('@' . $dateModifiedParsed);
-                } else {
-                    $dateStandards = [
-                        DateTime::RSS,
-                        DateTime::RFC822,
-                        DateTime::RFC2822,
-                        null,
-                    ];
-                    foreach ($dateStandards as $standard) {
-                        try {
-                            $date = date_create_from_format($standard, $dateModified);
-                            break;
-                        } catch (\Exception $e) {
-                            if ($standard === null) {
-                                throw new Exception\RuntimeException(
-                                    'Could not load date due to unrecognised format'
-                                    . ' (should follow RFC 822 or 2822): ' . $e->getMessage(),
-                                    0,
-                                    $e
-                                );
-                            }
-                        }
-                    }
+                try {
+                    $date = new DateTime($dateModified);
+                } catch (\Exception $e) {
+                    throw new Exception\RuntimeException(
+                        'Could not load date due to unrecognised format'
+                        . ' (should follow RFC 822 or 2822): ' . $e->getMessage(),
+                        0,
+                        $e
+                    );
                 }
             }
         }
--- a/src/Reader/Entry/Rss.php
+++ b/src/Reader/Entry/Rss.php
@@ -22,12 +22,10 @@ use stdClass;
 
 use function array_key_exists;
 use function count;
-use function date_create_from_format;
 use function in_array;
 use function is_array;
 use function is_string;
 use function preg_match;
-use function strtotime;
 use function trim;
 
 class Rss extends AbstractEntry implements EntryInterface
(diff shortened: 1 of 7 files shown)


froschdesign:
@bramley 
Please use the current release branch for bugfixes: 2.26x. More on the workflow can be found in the [contributing guidelines](https://github.com/laminas/.github/blob/a5f819c70aa3a7f91ab75a612314be8cbb240168/CONTRIBUTING.md#working-on-a-patch).

There some notes from Psalm which must be fixed: https://github.com/laminas/laminas-feed/actions/runs/23681249984/job/68993426854?pr=119#step:3:486

> Unit tests before and after the change have the same result
> 
> ```
> TESTS_LAMINAS_FEED_READER_ONLINE_ENABLED=1 phpunit --group Laminas_Feed --verbose
> 
> OK, but incomplete, skipped, or risky tests!
> Tests: 1536, Assertions: 1760, Skipped: 2, Incomplete: 11.
> ```

This can be seen from the checks in the continuous integration pipeline.

---

I am still taking a close look at the actual changes.
Thanks in advance for your help!

bramley:
I have rebased on 2.26.x.

froschdesign:
@bramley 
> I have rebased on 2.26.x.

Sorry, I suppose I did not make myself clear: The branch must be changed to 2.26.x because bugfix can be done via a patch release. The branch 2.27.x is for the next minor release which will be used for new features. This all is defined in semantic versioning: https://semver.org

bramley:
I have added unit tests for invalid feed last build date and invalid entry date modified to test that an exception is thrown.

bramley:
I don't know what to do about this.

In the test testGetLastBuildDateShouldThrowExceptionForInvalidDate()

    $feed->getLastBuildDate();

causes a psalm error about the return value of getLastBuildDate possbily being unused.

Changing it to 

    $date = $feed->getLastBuildDate();

causes an error about the variable $date being unused.

In the psalm documentation I then found advice to prefix the variable name with _

That seems to fix the psalm error but now causes a phpcs error

Error: Variable "$_date" is not in valid camel caps format

test/Reader/Feed/RssTest.php
```diff
@@ -2109,6 +2110,23 @@ public function testGetsLastBuildDateFromRss20None(): void
...
+            )
+        );
+
+        $this->expectException(Reader\Exception\RuntimeException::class);
+        $this->expectExceptionMessage(
+            'Could not load date due to unrecognised format (should follow RFC 822 or 2822)'
+        );
+
+        $this->assertInstanceOf(Rss::class, $feed);
+        $_date = $feed->getLastBuildDate();
```
froschdesign: The variable name is not allowed by the coding standard and only the call is needed here.

```suggestion
        $feed->getLastBuildDate();
```
Please remove the variable and we will see the result of the checks.
Thanks in advance! 👍🏻

froschdesign: @bramley 
Thanks! 👍🏻

src/Reader/Entry/Rss.php
```diff
@@ -206,31 +204,15 @@ public function getDateModified()
         ) {
             $dateModified = $this->xpath->evaluate('string(' . $this->xpathQueryRss . '/pubDate)');
```
froschdesign, line 205: It would be good to add a DocBlock with the types here, it should be `string|false`.

src/Reader/Feed/Rss.php
```diff
@@ -205,30 +204,13 @@ public function getDateModified()
                 $dateModified = $this->xpath->evaluate('string(/rss/channel/lastBuildDate)');
```
froschdesign, line 204: It would be good to add a DocBlock with the types here, it should be `string|false`.
[Then the baseline file of Psalm can be update](https://psalm.dev/docs/running_psalm/dealing_with_code_issues/#using-a-baseline-file), this should remove open problem with `PossiblyUnusedReturnValue`.

bramley:
I won't be doing any more work on this change.

froschdesign:
@bramley 
I'm sorry, I've been very busy at work and haven't been able to look into it further. But I'd be happy to take on your fix.
I will have a look at it in the near future and make any necessary changes myself.

PR #127, opened 2026-09-19
Slash: Don't Force Zero Comments onto RSS Items

johnchase-ca:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This is a bugfix for the Slash renderer that is exclusive to RSS 2.0 (the extension does not render anything on Atom feeds).

Presently, `Laminas\Feed\Writer\Extension\Slash\Renderer\Entry` is forcing a comment count of **zero (0)** onto every item that **does not have one set**:

```xml
<slash:comments>0</slash:comments>
```

This is improper because:

1. The element is un-requested and un-necessary (and therefore presumably also un-anticipated).
2. It implies support for comments in an item that doesn't implement them.
3. It will report inaccurate comment counts for items that do implement comments but don't report counts in RSS.
4. It forces the "slash" namespace (http://purl.org/rss/1.0/modules/slash/) onto every RSS feed, even if an application isn't intending to use it.

The final item can result in the W3C validator giving further recommendations to *"Ensure lastBuildDate is present when slash:comments is used"*, and is contrary to the design intentions of the renderer's own `$called` property, which per its comments is *"used to prevent premature appending of a XML namespace declaration until an element which requires it is actually appended"*.

My very small fix instructs the `_setCommentCount()` method to simply `return` rather than force a value of zero if no actual numeric comment count has been set.

N.B.- The Slash extension is still quite happy to output comment counts of zero (0) if that value **has been set**, but no longer makes them appear out of thin air.

I ran all unit tests on my local machine under PHP 8.3.6, and this modification created no new issues.


src/Writer/Extension/Slash/Renderer/Entry.php
```diff
@@ -61,8 +61,8 @@ protected function _appendNamespaces()
     protected function _setCommentCount(DOMDocument $dom, DOMElement $root)
     {
         $count = $this->getDataContainer()->getCommentCount();
-        if (! $count || ! is_numeric($count)) {
-            $count = 0;
+        if (! is_numeric($count)) {
```
froschdesign: This change is wrong, because the value of `<slash:comments>` [must be a positive integer](https://web.resource.org/rss/1.0/modules/slash/#model) and `0` is a positive integer. So the previous check is correct.

src/Writer/Extension/Slash/Renderer/Entry.php
```diff
@@ -61,8 +61,8 @@ protected function _appendNamespaces()
...
+            return;
```
froschdesign: Do we have a test for that?

froschdesign (changes requested):
Good catch!

src/Writer/Extension/Slash/Renderer/Entry.php
johnchase-ca: If we revert to the previous check, the FALSE check will reject explicitly set comment counts of zero (0), and prevent them from being rendered.

This would mean Laminas Feed is enforcing a stricter interpretation of the Slash specification than the standards bodies currently are.

I discovered in testing that although the Slash specification from ~25 years ago does define `slash:comments` as a _"positive integer"_, the current interpretation of both the RSS Board and W3C validators is that _"slash:comments must be a non-negative integer"_, and both consider 0 to be a valid value.

Additionally, it might escalate this PR to a BC Break, as systems that are currently generating valid RSS feeds with comment counts of 0 on new items would suddenly see those elements disappear and wonder why.

Please advise how you'd like me to proceed.

froschdesign: Damn, I specifically wanted to allow zero, but I misread the code. 🤦🏻‍♂️ 
You’re right, of course, because "0 comments" can obviously be a valid statement, so negative integers should be excluded.

johnchase-ca: Understood! 🙂

I believe we have achieved that, because `Laminas\Feed\Writer\Entry::setCommentCount()` only allows non-negative integers to be set.

So we have stopped forcing zero values, but the new code can continue to presume that any existing numeric value is valid and should be rendered.

src/Writer/Extension/Slash/Renderer/Entry.php
johnchase-ca: I found an existing `testCommentCountRendered()` test in `test/Writer/Renderer/Entry/RssTest.php` to check that a **positive** value gets rendered.

I have added `testCommentCountRendered0()` to check that a **zero (0)** value also gets rendered.  It passed successfully on my system under PHP 8.3.6.

I don't think checking for invalid values on the renderer side of things is necessary, as `test/Writer/EntryTest.php` already contains a whole slew of tests to check that `setCommentCount()` doesn't set an invalid value in the first place.

